PR #189, opened 2024-12-06
Add core isolation and change cores distribution

wmousa:


noctarius:
Please run `scripts/generate.sh` to generate cli.py and don't change it manually. It will remove the chance of running out of sync.

PR #231, opened 2025-03-17
Fix snapshot creation error handling

mxsrc:
The signature of `snapshot_controller.add` was inconsistent, returning either a tuple or a value. At the call-sites, this was not handled adequately.

This change consistently returns a tuple and process it adequately when calling the function.

## JIRA ticket link/s

-   [SFAM-1836](https://simplyblock.atlassian.net/browse/SFAM-1836)

## Summary of changes

1.  Make the signature of `snapshot_controller.add` consistent, always returning a tuple
2. Adapt call-sites to infer an error accordingly

### DIFF
--- a/simplyblock_cli/clibase.py
+++ b/simplyblock_cli/clibase.py
@@ -424,7 +424,9 @@ class CLIWrapperBase:
         new_size = 0
         if args.resize:
             new_size = self.parse_size(args.resize)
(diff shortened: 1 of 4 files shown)


mxsrc:
I noticed this also applies to `clone`, so I applied the same changes there.

Hamdy-khader (approved):

Hamdy-khader:
@mxsrc Approved and Merged. Thanks!

PR #232, opened 2025-03-17
Fix CLI parameter naming for snapshot creation

mxsrc:
The ID parameter for snapshot creation is named `snapshot_id`, instead of `volume_id`. The resulting snapshot ID is autogenerated and to set by the user.


noctarius:
Good catch!

PR #234, opened 2025-03-20
adjusted cluster deploy command

geoffrey1330:
## JIRA ticket link/s

-   [EXAMPLE-123](https://simplyblock.atlassian.net/browse/EXAMPLE-123)

## Summary of changes

1.  [ Provide some key summary of changes to give context as to why these changes have been made ]

## Additional Info

_[Any additional info such as screenshots or postman collections,etc that will help the reviewer review your changes]_


geoffrey1330:
> Just my two cents:
> 
> * `cli.py` is autogenerated and should not be modified directly, instead, modify the `cli-referance.yml` and run `./scripts/generate.sh`.
> * The PR includes modifications to `env_var` which should not be merged to the main branch.

Thanks on the review. 
for point 2 the env_var will be changed to reference `sbcli-dev` and `main` tag before merging

noctarius:
@geoffrey1330 What is the idea of the blacklist property? Just prevent it from bring generated? Why would I want to keep it in the yaml then?

geoffrey1330:
> @geoffrey1330 What is the idea of the blacklist property? Just prevent it from bring generated? Why would I want to keep it in the yaml then?

@noctarius We will be keeping it in the yaml cause we need their default value in this condition 
` if not self.developer_mode:`

noctarius:
I think something like `defaultOnly` might be a better term since blacklist sounds like "hide it completely from generation". Does that makes sense? Happy to quickly adjust it and add a bit more documentation to the property.

noctarius (approved):

geoffrey1330:
> @geoffrey1330 is this PR still relevant?

Yes it is @boddumanohar

geoffrey1330:
> Cool 🔥 . Could you please resolve the conflicts then we should be good to go 🚀

conflicts resolved

PR #238, opened 2025-03-24
Fix caching node nqn device

mxsrc:
## Summary of changes

Both `storage_node_ops` and `caching_node_controller` provide slightly different versions of `addNvmeDevices`. Recently, b9e69bb068eb6ca15fe24ecda12a5384e2488036 introduced a new flag to enumerate devices by NQN instead of serial number. This was only applied to one version. To fix this problem and avoid similar issues in the future, this PR:

1. unifies the different versions of `addNvmeDevices`
2. removes one and replaces it with calls to the version provided in `storage_node_ops`
3. removes co-located unused code

A change in semantics occurs because the caller's `rpc_client` is used which uses different timeout values.


Hamdy-khader (approved):
Looks good!
Thanks @mxsrc

PR #240, opened 2025-03-25
Added check to ensure cleanly generated cli.py

noctarius:
This PR enhances the builders (test and deploy) to check that a freshly generated cli.py is the same as the checked in, to prevent lost changes on follow up commits which regenerate cli.py and override manual changes.

### DIFF
--- a/.github/workflows/python-generate-test.yml
+++ b/.github/workflows/python-generate-test.yml
@@ -1,9 +1,6 @@
-name: Python Unit Testing
+name: Python Package Build
 
 on:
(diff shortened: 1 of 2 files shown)


noctarius:
This must be fixed after #241 as it'll otherwise fail the build.

mxsrc (commented):
Is there a way to avoid the duplication of the 'Checking cli.py' status command?
Ideally, I feel we'd have workflows execute one after the other: Test -> Build -> Publish.

PR #241, opened 2025-03-25
Fixed nargs on positional arguments which were ignored

noctarius:
Positional arguments (such as `volume_id` on `volume delete`) were ignored during generating cli.py. This PR fixes the problem and reintroduces the option to provide multiple volume ids.

### DIFF
--- a/scripts/cli-wrapper-gen.py
+++ b/scripts/cli-wrapper-gen.py
@@ -115,7 +115,8 @@ def get_description(item):
         return "<missing documentation>"
 
 
-def nargs(value):
+def nargs(item):
+    value = item["nargs"]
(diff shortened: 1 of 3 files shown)


mxsrc:
I don't exactly understand the change, the generated `cli.py` seems to be semantically identical. What are the semantic changes of passing looking up `nargs` from `parameter` in the filter instead of the tempalte. The removed shebang is the generation from a previous change?. Could you explain the change a bit more?

noctarius:
@mxsrc  The nargs in cli.py was manually added. That's why it looks similar. It was removed though when generate.sh was executed.

noctarius:
Oh, missed the removal of the shebang. Let me add this to the template.

mxsrc (approved):
Looks good to me :)

PR #242, opened 2025-03-25
Indicate minimum python version

mxsrc:
As discussed, this explicitly indicates the minimum python version we expect. 3.9 is the oldest non-EOL release and present Rocky 9.


setup.py
```diff
@@ -76,6 +76,7 @@ def get_requirements():
 setup(
     name=COMMAND_NAME,
     version=VERSION,
+    requires_python='>= 3.9',
```
Hamdy-khader, line 79: I suggest 3.7 as min.

mxsrc: Are there any systems we run on that use this? Iirc the consensus of the meeting yesterday was to use whatever Rocky Linux 9 ships, which is 3.9.

boddumanohar: https://github.com/simplyblock-io/sbcli/blob/e73abc3ade9b7c519c866fe5edbb329f72e6dcab/docker/Dockerfile_base#L1-L2

We've have been using python3.9 for our internal testing so far. So I'd agree with having 3.9 as the minimum python version.

boddumanohar (approved):

PR #244, opened 2025-03-26
Deduplicate helpers

mxsrc:
This PR unifies use of node helpers:
- the `run_command` helper in `node_utils` is removed and the one from `shell_utils` is used in its place
- names are changed to indicate them as public
- duplicate helpers in `caching_node_ops` are removed

### DIFF
--- a/simplyblock_core/shell_utils.py
+++ b/simplyblock_core/shell_utils.py
@@ -3,7 +3,10 @@ import subprocess
 
 
 def run_command(cmd):
-    process = subprocess.Popen(
-        cmd.split(), stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-    stdout, stderr = process.communicate()
-    return stdout.strip().decode("utf-8"), stderr.strip(), process.returncode
+    try:
+        process = subprocess.Popen(
+            cmd.split(), stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        stdout, stderr = process.communicate()
+        return stdout.strip().decode("utf-8"), stderr.strip(), process.returncode
+    except Exception as e:
+        return "", str(e), 1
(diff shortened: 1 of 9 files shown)


simplyblock_core/shell_utils.py
```diff
@@ -3,7 +3,10 @@
...
-    process = subprocess.Popen(
-        cmd.split(), stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-    stdout, stderr = process.communicate()
-    return stdout.strip().decode("utf-8"), stderr.strip(), process.returncode
+    try:
+        process = subprocess.Popen(
+            cmd.split(), stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        stdout, stderr = process.communicate()
+        return stdout.strip().decode("utf-8"), stderr.strip(), process.returncode
+    except Exception as e:
```
boddumanohar, line 11: Would it make sense to let the caller handle this exception instead?

mxsrc: Definitely, though we make little use of exception based error handling yet. It would make sense to change set, it simplifies lots of cumbersome error handling which is easy to get wrong. The `subprocess` module already has helpers that would be suitable for this, something like this would be what I'd prefer:
```
def run_command(command):
    return subprocess.check_output(command.split(' '), stderr=subprocess.PIPE, text=True)
```
It'd throw automatically including a descriptive exception, including the stderr output. We'd have to change all call-sites to handle this appropriately though. I didn't make that change since I thought avoiding exceptions was a conscious descision, if we want to use more of them I can adapt the PR.

mxsrc:
I found an additional duplicate helper, `get_nics_data` was present in both `simplyblock_core.utils` and `simplyblock_web.node_utils`.

Hamdy-khader (commented):
Looks good, one question thu, did you test those changes in a deployment or e2e ? @mxsrc  
can be installed using:
$pip install sbcli-pr244

mxsrc:
I did deploy a cluster using the branch and ran my API tests, not the e2e tests though, I have yet to learn how to run those.

PR #245, opened 2025-03-26
Remove dead code

mxsrc:
This PR includes multiple commits that prune unused code.

### DIFF
--- a/simplyblock_core/models/nvme_device.py
+++ b/simplyblock_core/models/nvme_device.py
@@ -44,19 +44,12 @@ class NVMeDevice(BaseModel):
     nvmf_ip: str = ""
     nvmf_nqn: str = ""
     nvmf_port: int = 0
-    overload_percentage: int = 0
-    partition_jm_bdev: str = ""
-    partition_jm_size: int = 0
-    partition_main_bdev: str = ""
-    partition_main_size: int = 0
-    partitions_count: int = 0
     pcie_address: str = ""
     physical_label: int = 0
     pt_bdev: str = ""
     qos_bdev: str = ""
     remote_bdev: str = ""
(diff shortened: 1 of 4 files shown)


simplyblock_core/models/nvme_device.py
```diff
@@ -46,19 +46,12 @@ class NVMeDevice(BaseModel):
     nvmf_ip: str = ""
     nvmf_nqn: str = ""
     nvmf_port: int = 0
-    overload_percentage: int = 0
-    partition_jm_bdev: str = ""
-    partition_jm_size: int = 0
-    partition_main_bdev: str = ""
-    partition_main_size: int = 0
-    partitions_count: int = 0
```
boddumanohar, line 54: removing this will also have to needs updating `simplyblock_web/static/swagger.yaml` 


Also these variables are properties within the FoundationDB. So I am not 100% confidant if removing this will preserve backward compatibility. 

So I leave out this variables and handle only the dead functions

boddumanohar (commented):
It's a good effort to remove dead code. But is this the exhaustive set of functions that are un-used or do you think there are more un-used functions?

simplyblock_core/models/nvme_device.py
mxsrc: I reintroduced the variables but marked them as unused.

mxsrc:
There is plenty of more unused code in the repository, this PR is what I stumbled upon during related changes. There are ways to test for this automatically, but not all can be found, e.g. functions since they may provided to downstream packages. Removing all dead code would be great, but there are already ~200 unused variables and imports so I didn't do an exhaustive refactor yet.

boddumanohar (approved):

PR #246, opened 2025-03-27
Various CLI changes

mxsrc:
This adapts the CLI to integrate some of the checks that are currently manual into the parsing itself, in particular adding types for:
- regexes
- sizes
- ranges

### DIFF
--- a/cli-reference-schema.yaml
+++ b/cli-reference-schema.yaml
@@ -110,7 +110,9 @@ properties:
                     - dest
                   not:
                     anyOf:
-                      - properties:  # private arguments cannot be required
+                      - description: Private arguments cannot be required
+                        x-intellij-html-description: Private arguments cannot be required
+                        properties:
                           private:
                             enum:
                               - true
@@ -118,12 +120,16 @@ properties:
                             enum:
                               - true
                         required: ["private", "required"]
-                      - properties:  # arguments with actions cannot be required
+                      - description: Arguments with actions cannot be required
+                        x-intellij-html-description: Arguments with actions cannot be required
+                        properties:
                           required:
                             enum:
                               - true
                         required: ["action", "required"]
-                      - properties:  # arguments with default values cannot be required
+                      - description: Arguments with default values cannot be required
+                        x-intellij-html-description: Arguments with default values cannot be required
+                        properties:
                           required:
                             enum:
                               - true
--- a/cli-reference-schema.yaml
+++ b/cli-reference-schema.yaml
@@ -108,6 +108,26 @@ properties:
                     - help
                     - type
                     - dest
+                  not:
+                    anyOf:
+                      - properties:  # private arguments cannot be required
+                          private:
+                            enum:
+                              - true
+                          required:
+                            enum:
+                              - true
+                        required: ["private", "required"]
+                      - properties:  # arguments with actions cannot be required
+                          required:
+                            enum:
+                              - true
+                        required: ["action", "required"]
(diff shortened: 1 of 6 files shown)


cli-reference-schema.yaml
```diff
@@ -165,25 +165,42 @@ properties:
...
-                        <ul>
-                        <li>str</li>
-                        <li>bool</li>
-                        <li>int</li>
-                        </ul>
-                      enum:
-                        - str
-                        - bool
-                        - int
+                      oneOf:
```
noctarius, line 168: We should reintroduce the documentation to show up in the IDEs.

cli-reference-schema.yaml
```diff
@@ -165,25 +165,42 @@ properties:
...
+                          properties:
+                            regex:
+                              type: string
+                        - type: object
+                          additionalProperties: false
+                          minProperties: 1
+                          properties:
+                            size:
+                              type: object
+                              properties:
```
noctarius, line 199: I feel like min and max on the size type should have regex patterns applied to match the string that can be set as a size descriptor

scripts/cli-wrapper-gen.py
```diff
@@ -131,7 +141,7 @@ def nargs(item):
             for subcommand in command["subcommands"]:
                 if "arguments" in subcommand:
                     for argument in subcommand["arguments"]:
-                        argument["required"] = False if "default" not in argument else True
+                        argument["required"] = argument["required"] if 'required' in argument else ("default" in argument)
```
noctarius, line 144: Shouldn't the latter part be `"default" not in argument` to make is optional if it has a default? 🧐

noctarius (changes requested):

noctarius:
The cli.py isn't up to date :P

cli-reference-schema.yaml
```diff
@@ -165,25 +165,66 @@ properties:
...
-                        The parameter type of the positional argument or parameter.<br>
-                        Valid options are:<br>
+                        The data type of the argument. The following options are allowed:<br>
                         <ul>
-                        <li>str</li>
-                        <li>bool</li>
-                        <li>int</li>
+                        <li>'str', 'int', 'bool'</li>
+                        <li>'regex': For strings validated against a regex. E.g.: `{"regex": "^foo$"}`</li>
+                        <li>'size': For size types. Any integer is valid, along strings including SI/IEC suffixes.
```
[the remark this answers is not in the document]
noctarius: Same in the version above

mxsrc:
I changed the docs again and rebased onto main. The diff was from an unclean state in the main version I was based on.

noctarius (approved):
LGTM

But I think we shouldn't merge it into 2025.3 anymore. Should go into master after the branch is created. Or did that happen already?

mxsrc:
Alright, then let's hold of until then.

cli-reference-schema.yaml
```diff
@@ -165,25 +191,72 @@ properties:
...
+                          x-intellij-html-description: |
+                            Simple data types representing either integer, string, boolean, or data size.
+                          enum:
+                            - int
+                            - str
+                            - bool
+                            - size
+                        - type: object
+                          description: |
+                            Valid arguments are integers within a specific range. Minimum and maximum are inclusive.
```
noctarius, line 213: Just realized, maximum is exclusive.

mxsrc: True, fixed.

boddumanohar:
will hold merging of this PR until after the release

PR #247, opened 2025-03-27
Address pylint errors

mxsrc:
This fixes errors pylint highlighted.


simplyblock_core/services/spdk_http_proxy_server.py
Hamdy-khader, line 1: remove "cert=None" from line 134

mxsrc: Good catch, done :)

PR #248, opened 2025-03-28
Added schema check of the cli-reference.yaml

noctarius:
This PR introduces schema validation against the `cli-reference.yaml` to prevent illegal or misused properties.

If errors are found, the validation errors are listed on stderr for easy analysis of the schema validation failure and fixing.

Example output:
```
(.venv) demo@demo ~/s/sbcli > scripts/generate.sh
Generator failed on schema validation. Found the following errors:
 - $.commands[1].subcommands[0].arguments[33].type: 1 is not of type 'string'
 - $.commands[1].subcommands[0].arguments[33].type: 1 is not one of ['str', 'bool', 'int']
```

### DIFF
--- a/scripts/cli-wrapper-gen.py
+++ b/scripts/cli-wrapper-gen.py
@@ -3,6 +3,10 @@ import yaml
 import sys
 import re
 
+from jsonschema import validators
+from jsonschema.exceptions import ValidationError
+
+
 def is_parameter(item):
     return item["name"].startswith("--") or item["name"].startswith("-")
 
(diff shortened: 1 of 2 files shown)


mxsrc (approved):
Looks good to me! Since you asked about the iteration, I'd avoid explicit iteration wherever possible:

            errors = list(validator.iter_errors(reference))

            if errors:
                print("Generator failed on schema validation. Found the following errors:", file=sys.stderr)
                print('\n'.join(f" - {error.json_path}: {error.message}" for error in errors), file=sys.stderr)
                sys.exit(1)

PR #254, opened 2025-04-03
Renamed --node-ip to --node-addr

noctarius:
This PR renamed the `storage-node restart --node-ip` parameter to `--node-addr` as it requires the port number to be given as value (e.g. `192.168.0.1:5000`).

The `--node-ip` parameter name still works as it is added as an alias. It should be marked as deprecated and eventually removed.

This PR is related to: https://github.com/simplyblock-io/documentation/issues/32

### DIFF
--- a/cli-reference.yaml
+++ b/cli-reference.yaml
@@ -262,12 +262,15 @@ commands:
             type: str
             default: "0"
             private: true
-          - name: "--node-ip"
+          - name: "--node-addr"
+            aliases:
+              - "--node-ip"
             help: "Restart Node on new node"
             description: >
               Allows to restart an existing storage node on new host or hardware. Devices attached to storage nodes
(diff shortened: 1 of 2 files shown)


boddumanohar (approved):
https://github.com/simplyblock-io/sbcli/blob/aec083fa5aa66e50426874121b9b612622f42dc6/simplyblock_cli/clibase.py#L143-L144


should we also update its update its usage to `args.node_addr`?

noctarius:
> https://github.com/simplyblock-io/sbcli/blob/aec083fa5aa66e50426874121b9b612622f42dc6/simplyblock_cli/clibase.py#L143-L144
> 
> should we also update its update its usage to `args.node_addr`?

In theory yes, but since it's wrong internally all the way, I didn't want to touch it for now.

noctarius:
Rebased to main

PR #265, opened 2025-04-07
Patched handling of default ndcs + npcs for `cluster add` and `cluster create`

deverts:
Our internal deploy tooling got stung by this. I made a fix to the default parameters, and changed behavior for when 0x0 is provided. Previously, it then converted this to a 2x1, I've made this error instead. My thinking is that if a user provides 0x0, it's probably an error.

In cluster add:
- set defaults to 1x1
- error if the user sets these both to 0 (previously: it would default to 2x1)

In cluster create:
- change to error if the user provides 0x0 (prev: it would default to 2x1)

I split this out into two commits in case you only want the first change.

### DIFF
--- a/cli-reference.yaml
+++ b/cli-reference.yaml
@@ -1199,12 +1199,12 @@ commands:
             help: "Erasure coding schema parameter k (distributed raid), default: 1"
             dest: distr_ndcs
             type: int
-            default: 0
+            default: 1
           - name: "--parity-chunks-per-stripe"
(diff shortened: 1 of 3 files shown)


noctarius (approved):
Thanks @deverts!

Looks good to me.

@Hamdy-khader can you also have a quick look?

noctarius:
Thanks @deverts! Nice find with the inconsistency of the documentation and code 👍

PR #266, opened 2025-04-08
Fix SFAM-1921

mxsrc:
## JIRA ticket link/s

- [SFAM-1921](https://simplyblock.atlassian.net/browse/SFAM-1921)

## Summary of changes
The `/spdk_process_start` endpoint is internal API and expects all sizes to be passed as integers, i.e. without units.
- Removed call to `parse_size` on `total_mem` in `snode_ops.spdk_process_start`
- Adapted `parse_size` to be more robust when passed an integer.


boddumanohar (approved):
As per the comments on the tickets, it looks like this has already been validated.

simplyblock_core/utils.py
```diff
@@ -629,12 +629,8 @@ def parse_size(size_string: str, mode: str = 'si/iec', unit: str = '', strict: b
...
-        try:
-            x = int(size_string)
-            return x
-        except Exception:
-            pass
+    if isinstance(size_string, int):
+        return size_string
 
     try:
         m = re.match(r'^(?P<size_in_unit>\d+) ?(?P<unit>\w+)?$', size_string.strip())
```
Hamdy-khader, line 636: Replace in same file line 654 with:

```
base, exponent = _parse_unit(unit, 'si/iec'
size = parse_size(size)
```

Hamdy-khader (changes requested):
Looks good, please fix my commit then I will merge

mxsrc:
I spotted a condition I missed, passing an integer while assuming a unit was not handled correctly. The updated version processes this as expected. I also added test cases that trigger the situation present in SFAM-1921 to ensure this is not missed in the future.

Hamdy-khader (approved):
Looks good, Merging!
Thanks @mxsrc

Hamdy-khader:
Merged to version 17.4.25

PR #237, opened 2025-03-24
Fix webapp inconsistency

mxsrc:
## Summary of changes
413d558086fc0b81754db1f0398115c54b819124 introduced `simplyblock_web/node_webapp.py`, replacing `simplyblock_web/{snode,caching_node}_app.py`. 5657da8923e35101d27a2ce979aadfe2330412f7 reverted the use of the new entrypoint for the caching node API

1. a29ae30 Removes the left-over `snode_app{,_k8s}.py`
2. 43c9b4c Removes `caching_node{,_k8s}.py` and changes its usages to the common entrypoint.

If the second change is not desired, I feel the entry point should be adapted to remove the alternative `caching_node` usage.

PR #267, opened 2025-04-09
Use flask_openapi3 for internal API

mxsrc:
As discussed, this PR proposes to use flask_openapi3 for our internal API. The annotations are more elaborate, but enable automated validation of passed parameters, and provide generated documentation/specification of our API. This may be used to automatically confirm consistency of our clients with the servers, and reduces the effort spent on manual validation.

For now the PR is a draft, I did some preliminary tests that confirm basic functionality. All of our internal APIs are transcribed in this, I only added documentation beyond semantics for the basic node API to showcase how that would look. The functions' docstrings are used for the endpoint's `description`, the `summary` is passed to the annotation.

What do you think @Hamdy-khader , @noctarius 

P.S.: The PR is based on #237 , that may be removed as well but it makes sense to address in one go.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,3 +20,4 @@ prometheus-client
 ec2-metadata
 flask-swagger-ui
 sentry-sdk[flask]
+flask-openapi3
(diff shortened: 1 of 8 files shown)


mxsrc:
After discussion with Michael we think this should be merged after the 25.3 release as it's a major change. After we gained some experience and confidence we can create a similar changeset for the external API.

PR #276, opened 2025-04-18
Renamed parameter still referred to with old name in error message

noctarius:
We renamed `--max-prov` to `--max-size`. The error message, however, still uses the old name. This PR fixes the error message.


mxsrc (approved):
lgtm. Will be superseeded though by #246 once that's merged.

noctarius:
Agreed, but I think we won't merge 246 before the final release as far as I remember.

mxsrc:
True!

PR #284, opened 2025-04-23
integrate ruff linter

boddumanohar:
## Summary of changes

1. integrate linter to identify syntax errors, missing imports, circular imports, incorrect variables in pull requests.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,0 +1,17 @@
+# .github/workflows/lint.yml
+name: Python Linting
+
+on:
+  push:
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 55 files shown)


noctarius:
As a generated file cli.py needs to be excluded, otherwise the build check for the generator will fail

mxsrc:
I like this! The PR currently still includes changes to `cli.py` that seem to be invalid.

boddumanohar:
Thanks for the feedback guys. Will address them. Since we are nearing release, I don't want to push the changes as a part of the current release.

mxsrc:
The added tox configuration refers to environments that are not present in the configuration. I'd suggest removing the `type`, and `3.X` entries from `tox.ini`.

PR #287, opened 2025-04-24
Add basic integration tests

boddumanohar:
## Summary of changes

Added sample integration tests file that runs for every PR and merge `main` 
There is still lot of scope of improvement here. These tests are just added a simple check

## Additional Info

Sample output: 

Screenshot 2025-04-24 at 20 03 31


noctarius:
I think the integration test should test more than just the return code. While it's a start, not sure how much worth just checking the return code is. Did you test that it actually fails with a broken call?

We should probably also add the same set of tests via the CLI.

boddumanohar:
yes definitely a lot of room for improvement. 

The ones which is commented out fails at the moment. Hence the failure scenario is also tested

Screenshot 2025-04-25 at 11 51 28

CLI and API both calls same code that's present in `simplyblock_core`. I just went with API as it's much faster to work with because they return JSON

noctarius:
Every start is a good start! LGTM

noctarius (approved):

mxsrc (approved):
I like this! A few comments:
- I'm wondering how much of the system can be tested without actually deploying a cluster, but doing that will be much more costly from a CI action
- The actual tests would be more easily adapted if we'd use a testing framework like pytest, but we can adapt this at a later point if that is a shared opinion
- `check_endpoints.py:52` is commented out without further information, can that be removed

boddumanohar:
I believe a lot of APIs can be tested without actually deploying the entire system. considering the challenges of the system, I think we'll have to draft an approach how we plan to implement integration tests. 

Thanks, regarding `check_endpoints.py:52` added a small comment

PR #288, opened 2025-04-24
Fix false-negative type checks

mxsrc:
The name collision between a module and instance caused the type checker to complain.

## JIRA ticket link/s

-   [SFAM-1967](https://simplyblock.atlassian.net/browse/SFAM-1967)

## Summary of changes
I chose the shorter name `db` for the actual `DBController` instance.


boddumanohar (approved):
Observed this when running the changes through the linter. Considering the size of the changes, having a separate PR makes sense.

PR #289, opened 2025-04-24
Fix argument reference

geoffrey1330:
## Summary of changes

1.  fixed wrong reference of  method `parse_size`


mxsrc (approved):
Looks good to me! The changed lines could also be removed if I'm not mistaken, but this will fix the issue.

PR #292, opened 2025-04-25
Fix missing requirement

mxsrc:
`ec2_metadata` is used but not listed in requirements.txt.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,3 +19,4 @@ argcomplete
 jc
 sentry-sdk
 prometheus_client
+ec2_metadata


boddumanohar (approved):
Good find. But this package already exists in our simplyblock images because, we are using a base image. 

https://github.com/simplyblock-io/sbcli/blob/895933666d8aa73b1ab6f999b8c39785167f576c/docker/Dockerfile#L2


Screenshot 2025-04-25 at 16 22 33

probably we had it in the past and was somehow removed?

PR #319, opened 2025-05-12
priorize env variable over env_var file

boddumanohar:
When reading for env variables, prioritise env variables over `env_var`

### DIFF
--- a/simplyblock_core/constants.py
+++ b/simplyblock_core/constants.py
@@ -4,8 +4,13 @@ SCRIPT_PATH = os.path.dirname(os.path.realpath(__file__))
 
 
 def get_from_env_var_file(name, default=None):
+    """
+    priority of env var file is higher than env var
+    """
     if not name:
         return False
+    if os.env.get(name):
+        return os.env.get(name)
     with open(f"{SCRIPT_PATH}/env_var", "r", encoding="utf-8") as fh:
         for line in fh.readlines():
             if line.startswith(name):


mxsrc:
I agree with the idea of the change, but wonder whether this is the right approach. The function that is modified is named `get_from_env_var_file`, after this change, that would no longer be correct.

Thinking about this, I think a different appraoch would be to
1. on startup load the `env_var` contents into environment variables, unless their values are present
2. when using the variables, simply look up the values in the environment.

This is of course a bit more elaborate of a change, but I feel this approach would make things more intuitive and less dependent on on-disk state. Also, it allows to override the values from the env_var file without modifying it, as this PR intends.

boddumanohar:
The primary intention of this change is to have a way to override the existing configuration defined in `env_var` file rather than editing the file. There by prioritising env vars over config variable. 

so instead of the making the change in `get_from_env_var_file` function do you suggest to create a new function probably `get_config_var` which try to looking for a variable in env it doesn't exist, the fall backs to the current `get_from_env_var_file`?

mxsrc:
Exactly, at module import we could run a function that establishes the variables present in `env_var` in `os.environ`, unless they are present, and then either use those values in `constants.py`, e.g.:
```
SIMPLY_BLOCK_DOCKER_IMAGE = os.environ['SIMPLY_BLOCK_DOCKER_IMAGE']
```
or replace uses of `constants.SIMPLY_BLOCK_DOCKER_IMAGE` with `os.environ['SIMPLY_BLOCK_DOCKER_IMAGE']`.

mxsrc (approved):
Looks good to me. Thanks! It seems the integration tests fails, though without many details.

PR #326, opened 2025-05-14
Unify requirements files

mxsrc:
This merges the simplyblock_web/requirements.txt file into the top-level one. The project is packaged as one, so splitting up requirement files duplicates entries and invites inconsistencies.


requirements.txt
```diff
@@ -18,6 +17,7 @@ jinja2
 argcomplete
 jc
 sentry-sdk
-prometheus_client
-ec2_metadata
-
+prometheus-client
```
boddumanohar, line 20: Should we consider locking the versions so that builds are always reproducible and don't get affected by version incompatibilities?

boddumanohar (approved):

PR #337, opened 2025-05-23
Alceml create helper

mxsrc:
This introduces a helper for alceml device creation which is being used at different locations in the code and largely the same.


Hamdy-khader (approved):
+1

PR #338, opened 2025-05-26
use dockerhub images

boddumanohar:
## Summary of changes

Currently pulling the images for AWS ECR results in error

```
docker pull public.ecr.aws/simply-block/simplyblock:main
Error response from daemon: toomanyrequests: Data limit exceeded
```

This is due to 500GB/Month limit on for un authenticated pulls and this cannot be changed. 

https://aws.amazon.com/ecr/pricing/

Currently for docker hub images, for internal testing purposes testing, we could use our docker proxy. And in general, anyone could setup one quite easily using https://hub.docker.com/_/registry


mxsrc:
This occurence needs to be adapted as well: https://github.com/simplyblock-io/sbcli/actions/runs/15252403118/workflow?pr=338#L57

boddumanohar:
the integration tests are failing due to the same reason that this PR is trying to fix: 

```
toomanyrequests: Data limit exceeded
```

The tests, should be fixed when the PR is merged.

boddumanohar:
the value thats passes as a part of SBCLI_INSTALL_SOURCE is always getting over written. we should SBCLI_BRANCH instead. So pushed a change for the same.
https://github.com/simplyblock-io/sbcli/pull/338/commits/87d352c42b3bbe1d30a80b459982f53b98052346

The build for this PR will fail, because GitHub actions is taking the change for main branch

mxsrc (approved):
Looks good!

PR #341, opened 2025-05-27
Exchanged term node-ip for node-addr in storage node add

noctarius:
When adding a storage node, the node address (<ip>:5000) has to be given, not the node ip by itself. This PR changes the terminology and internal variable, as well as optimizes the help text.


mxsrc (approved):
Looks good!

PR #345, opened 2025-05-27
Introduce exception based error handling

mxsrc:
This PR introduces exception-based error handling. It lays the groundwork to start using exceptions more actively.

It adds contribution guidelines describing how to handle errors, and guards for uncaught exceptions at the entrypoints.
In the CLI, exceptions are caught and their message is printed. If the debug flag is set, the stack trace is printed as well
In the API, the exception is converted to a json representation and returned. If no explicit status code is set, 500 (Internal server error) is used.

The `cluster_ops` module is refactored to use exception based error handling.
In the process, a number of additional error checks were added, and some were removed as they were used to guard against exceptions being thrown later on.
Since the rest of the code-base is not yet converted to exceptions, many exceptions originate in this module now. With more of the code being transformed, these checks will be reduced. A common pattern is

```
ret = call()
if not ret:
    logger.error('msg')
    return False
```
which now often reads
```
ret = call()
if not ret:
    raise ValueError('msg')
```
and in the future will be shortened to
```
call()
```
as the exception originates in the `call` function.

In the process, I moved some of the output formatting out of the `cluster_ops` functions and into `clibase`.

Since the CLI expects all called functions to return a value that evaluates is truethy (`bool(ret) == True`), after calling the functions an unconditional `return True` is used, as errors are communicated via exceptions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,55 @@
+# Contribution guidelines
+
+## Error Handling Guidelines
+All contributions to this repository must follow exception-based error handling practices. This applies to:
+
+- **New code**: All newly introduced functions, methods, and modules
+- **Modified code**: Any existing code that is touched or refactored as part of your changeset
+
+When modifying existing code that uses other error handling patterns:
+
+- Convert the touched code sections to use exceptions
+- Update calling code within the same changeset if necessary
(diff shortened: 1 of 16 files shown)


Hamdy-khader:
@mxsrc Let's resolve conflicts and merge.
Thanks!

PR #348, opened 2025-05-28
Print response content on failed request

mxsrc:
Add some output to the integration test in case a request fails.


Hamdy-khader (approved):
+1

Hamdy-khader:
@mxsrc Thanks!

PR #365, opened 2025-06-03
[SFAM-2089] Missing get-logs cli parameter

mxsrc:


noctarius:
I think someone removed it a while back but I agree, it should be there 😁

mxsrc:
It was my bad, in #345 I moved the output formatting out of the `cluster_ops` function but forgot to introduce this parameter.

PR #384, opened 2025-06-11
Convert caching node ops to openapi generator

mxsrc:
The caching node API's implementation was duplicated and only partly converted to the new API framework introduced in #267. This PR removes the duplication and converts the remaining API.


noctarius:
From what I can see, it looks good.

PR #386, opened 2025-06-13
set up ruff linter for simplyblock_web and simplyblock_cli folders

boddumanohar:
## Summary of changes

fix lint issues in simplyblock_web and simplyblock_cli folders.


mxsrc (approved):
Great addition! Thanks :)

PR #387, opened 2025-06-13
setup mypy int in simplyblock_web and simplyblock_cli folders

boddumanohar:
## Summary of changes

setup mypy int in simplyblock_web and simplyblock_cli folders


mxsrc (commented):
Thanks for the PR, I think it's a great idea to apply the type checker more thoroughly.

I'm wondering why were the imports in the API tests changed? pytest code is not (should not be) part of the overall project, it expects to be executed from anywhere and so using relative imports is expected. This also avoids the `PYTHONPATH` modification in `api-tests.yml`

Also, I hesitate regarding the PR. I agree with the purpose, but as is, mypy by default only checks functions called from withing functions with type definitions. Using the `check_untyped_defs` flags, our project at the moment has ~500 errors which should be addressed. If we merge this PR, I fear that people will add annotations, which is a good idea, but consequently run the checker against a wider part of the code base, which might prompt people to not use type annotations, instead of fixing them.

For that reason I'm wondering whether it might do more harm than good in the end, I'm not sure about that though.
Alternatively, we might apply the changes, but just not add `mypy` to the CI _yet_. Or we include it in the CI, but run it against the whole codebase and do not consider its failure a workflow failure, for now.

What do you think?

boddumanohar:
currently tests run for pytest are inside simplyblock_web folder. Due to this reason the linter detected issues.

You might be right for the usage of annotations perspective. But the changes in the PR will act as a good guard rail against accidental type issues. Where the devs use annotations or not can be provided as a feedback when the PRs are raised. But wouldn't want to identify issues during run time. To protect from such type issues could be an other notation to have.

mxsrc:
You're right, I agree this is a helpful addition to avoid type issues in the future. I'd like to exclude the tests from its execution though s.t. they can be used from contexts other than the `simplyblock_web` package.

PR #389, opened 2025-06-13
Consolidate python checks into single workflow

mxsrc:
We apply different checks to our python codebase, which are currently scattered across multiple workflows. This PR consolidates all checks into a single workflow. This improves discoverability, readability of check results, ensures they are applied consistently, and avoids duplication.

There is still some duplication in `python-publish.yml`, and `release.yml`, but as the release process is to be reworked anyway I left them untouched in this PR.


boddumanohar (approved):
Looks clean!. Thanks.

PR #391, opened 2025-06-13
fix lint errors of type Item None of StorageNode

boddumanohar:
## Summary of changes

this PR solves all the lint errors of type 
```
Item "None" of "StorageNode
```


mxsrc (approved):
I like it. I think in the long run the functions should raise `KeyError`s themselves on missing entities, but with the current implementation this is a good improvement.

PR #393, opened 2025-06-13
Apply safe ruff fixes to simplyblock_core

mxsrc:
Currently we apply the linter only to simplyblock_web and simplyblock_cli as simplyblock_core does not pass the checks and there are violations requiring non-trivial fixes. This PR applies the trivial, automatic ruff fixes to pave the way to using the linter on the core codebase as well.


mxsrc:
Haha, duplicate of #392 :)

PR #399, opened 2025-06-15
fixes sfam-2103

Hamdy-khader:
## JIRA ticket

-   [SFAM-2103](https://simplyblock.atlassian.net/browse/SFAM-2103)

## Changes

1.  check node health conditions before port unblock
2. replicate journal explicitly on restart based on conditions of secondary


mxsrc (approved):
Could you give a short rationale of the changes, e.g. why was the previous cleanup not working and the asynchronous process is required? And what necessitates the explicit synchronization? Maybe in that case a comment would help other readers understand the thought process.
Just based on the code this looks good, but I cannot give any judgement on implications

PR #400, opened 2025-06-15
Fixes sfam-2093

Hamdy-khader:
## JIRA ticket link/s

-   [EXAMPLE-123](https://simplyblock.atlassian.net/browse/EXAMPLE-123)

## Summary of changes

1.  [ Provide some key summary of changes to give context as to why these changes have been made ]

## Additional Info

_[Any additional info such as screenshots or postman collections,etc that will help the reviewer review your changes]_


mxsrc (changes requested):
Is there a specific reason for commenting out the code instead of removing it? If so, could we document that in the code, otherwise I'm afraid it will get lost. If not, can we instead remove the blocks instead of commenting them? We can still recover them from git if need be.

mxsrc (approved):
Thanks, looks good!

PR #405, opened 2025-06-16
Raise error on missing storage node

mxsrc:
This changes the `DBController` API to raise an error on missing storage node IDs instead of returning None. This (and similar functions) are a large contributor to failing type checks, this PR reduces the `mypy` error count from 830 to 483.


schmidt-scaled (approved):
looks good to me. why can we remove the `sec_node is None`? you capture this in an exception instead?

PR #408, opened 2025-06-17
K8s mgmt service

geoffrey1330:
- Moved Mgmt Services to Kubernetes
- Introduced HA mongoDB, Prometheus, Opensearch 
- Update check for mgmt node to support both docker and kubernetes


noctarius:
In general, I feel like, we should depend more on the vendor's provided helm charts like https://github.com/mongodb/helm-charts/tree/main/charts for mongodb. In this case, we'd install the mongodb operator and CRDs (if not yet available) and just create the necessary CRD instance. If someone uses the enterprise version, that's fine, we could just use it. No need to install our own.

Same goes for things like OpenSearch, Prometheus, and others.

simplyblock_core/constants.py
```diff
@@ -151,3 +151,87 @@ def get_config_var(name, default=None):
 SYSTEM_INFO_FILE = "/etc/simplyblock/system_info"
 
 LVO_MAX_NAMESPACES_PER_SUBSYS=32
+
+K8S_NAMESPACE = "simplyblock"
```
boddumanohar, line 155: it would be great if users could have a way to set their own namespace

boddumanohar (approved):
The changes in this PR have coupled monitoring stack and simplyblock management stack. But for the day to day operation of Simplyblock mgmt node, monitoring stack is not needed and not dependent. So It makes sense have their deployments separate from monitoring deployments. 

I understand that having them coupled could provide a unified experience and helps  us deploy a production ready cluster in a few steps. 

In the further set of PRs related to k8s mgmt service, I'd try to decouple them. So the customer could have better control over monitoring.  And we should provide our suggested monitoring setup separately by taking the original operators as a base.

But for now so far this is really great effort and a good way to forward 🚀

noctarius:
Uhh nice, it's merged 🙌

PR #415, opened 2025-06-20
await for lint success before integration tests

boddumanohar:


mxsrc:
Just a heads-up: the changes in [1e03595](https://github.com/simplyblock-io/sbcli/pull/415/commits/1e035955ee8d5dc6ba68932f2acf2040ffffd056) will cause issues, the current database implementation uses type-hints to deduce the values and does not interpret `Optional` correctly.

PR #417, opened 2025-06-20
added k8s Job for kubelet restart

geoffrey1330:
- Added an initContainer to restart kubelet


simplyblock_web/templates/storage_deploy_spdk.yaml.j2
```diff
@@ -86,6 +86,26 @@ spec:
...
+      args:
+        - |
+          set -e
+          set -x
+
+          # Enter the host's mount namespace to ensure access to host binaries and files
+          nsenter --target 1 --mount --uts --ipc --net --pid -- /bin/bash -c '
+            if command -v systemctl >/dev/null 2>&1; then
+              echo "Restarting kubelet..."
+              systemctl restart kubelet && echo "Kubelet restarted successfully" || echo "Failed to restart kubelet"
```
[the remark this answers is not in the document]
geoffrey1330: Claimed by the kubelet or spdk?

simplyblock_core/env_var
```diff
@@ -1,6 +1,6 @@
-SIMPLY_BLOCK_COMMAND_NAME=sbcli-dev
-SIMPLY_BLOCK_VERSION=19.0.2
+SIMPLY_BLOCK_COMMAND_NAME=sbcli-mig
```
boddumanohar, line 1: updating this will fail the integration tests build. You could always install the sbcli from a specific branch using: 

```
pip install --upgrade git+https://github.com/simplyblock/sbcli.git@add_kubelet_restart_initContainer
```

and use sbcli-dev

simplyblock_web/templates/storage_deploy_spdk.yaml.j2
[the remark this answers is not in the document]
boddumanohar: yes, better having it separately

simplyblock_core/utils.py
```diff
@@ -1607,6 +1608,14 @@ def set_hugepages_if_needed(node, hugepages_needed, page_size_kb=2048):
...
         logger.error(f"Node {node}: Error occurred: {e}")
 
+def adjust_hugepages(hugepages):
+    remainder = hugepages % 500
+    hugepages =  hugepages + (500 - remainder)
+
+    str_val = str(hugepages)
+    decimal_val = float(str_val[0] + '.' + str_val[1])
+    add_val = int(decimal_val * 24)
+    return hugepages + add_val
```
boddumanohar, line 1618: Its a bit unclear about whats happening here. There seems like a lot of magic numbers used. Why 500? why 24?

maybe you could add comments to function explaining whats happing.

simplyblock_web/templates/storage_init_job.yaml.j2
```diff
@@ -0,0 +1,54 @@
...
+              set -e
+
+              echo "--- Installing curl & networking tools ---"
+              apt-get update && apt-get install -y iproute2 curl grep
+
+              NODE_IP=$(ip route get 1.1.1.1 | grep -oE 'src [0-9.]+' | awk '{print $2}')
+              echo "Detected node IP: $NODE_IP"
+
+              echo "--- Sending config to $NODE_IP ---"
+              RESPONSE=$(curl -s -o /dev/null -w "%{http_code}" -X POST http://$NODE_IP:5000/snode/apply_config -H "Content-Type: application/json" -d '{}')
```
boddumanohar, line 37: does `snode/apply_config` also applies the huge pages configuration? 

```
sysctl -w vm.nr_hugepages=$hugepages
```

if so since we are already doing everything. Should we ask user to configure the huge pages any more as a part of. node preparation steps? 

https://github.com/simplyblock-io/simplyblock-csi/blob/master/docs/storage-nodes.md#preparing-nodes

simplyblock_web/blueprints/snode_ops_k8s.py
```diff
@@ -358,6 +358,29 @@ def spdk_process_start(body: SPDKParams):
...
+        job_yaml = yaml.safe_load(job_template.render(values))
+        batch_v1 = core_utils.get_k8s_batch_client()
+        job_resp = batch_v1.create_namespaced_job(namespace=namespace, body=job_yaml)
+        msg = f"Job created: '{job_resp.metadata.name}' in namespace '{namespace}"
+        logger.info(msg)
+
+        node_utils_k8s.wait_for_job_completion(job_resp.metadata.name, namespace)
+        logger.info(f"Job '{job_resp.metadata.name}' completed successfully")
+
+        batch_v1.delete_namespaced_job(
```
boddumanohar, line 372: Maybe we could leave the job as it is without deleting it. That way the users could check the logs

simplyblock_core/utils.py
geoffrey1330: What happening here is we calculate the required hugpeage needed for spdk and the same time we need to set this hugepage in k8s pod Request and Limit and they are represented in Gi. 
In a situation spdk request **4966** of hugepage we need to 5Gi as the request and Limit and the representation of 5Gi is **5120**  thus the rounding `5000 + 5.0 * 24`

boddumanohar: thanks, can you please update the doc string of the func

PR #420, opened 2025-06-20
Address type check failures in json rpc client

mxsrc:


simplyblock_core/services/spdk/client.py
```diff
@@ -265,7 +273,7 @@ class GoClientResponse(ctypes.Structure):
             raise JSONRPCException(rpc_error)
 
         try:
-            json_resp = json.loads(ctypes.c_char_p.from_buffer(resp.response).value)
+            json_resp = json.loads(ctypes.c_char_p.from_buffer(resp.response).value)  # type: ignore
```
boddumanohar, line 276: does this also fail due to the db model type inference issue?

boddumanohar (approved):

simplyblock_core/services/spdk/client.py
mxsrc: No, this is a different issue. Apparently the value can technically be None. Since this has never been an issue and it is a very low-level call I opted to ignore it instead of explicitly addressing it.

PR #426, opened 2025-06-22
Fix huge page

wmousa:


Hamdy-khader:
added here https://github.com/simplyblock-io/sbcli/commit/1444fad666545320f55a7de8c868f889c1504bdc

PR #454, opened 2025-06-26
enable k8s for bare metal deploy

boddumanohar:
Modified bare-metal-deploy.yml to support deploying k8s clusters

Currently these changes only include to `bare-metal-deploy.yml` file. In the next PR integration with CI and with PRs follow.

### DIFF
--- a/.github/workflows/bare-metal-deploy.yml
+++ b/.github/workflows/bare-metal-deploy.yml
@@ -39,7 +39,7 @@ on:
       k8s_snode:
         description: "Run snode on k8s"
         required: false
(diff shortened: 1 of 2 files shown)


mxsrc (approved):
Looks good, if my nitpick is correct we might want to address that.

In the future I'd like to have this process elsewhere, simply called from here s.t. it can be used manually or in other integrations as well, but this is a good addition.

Also, we could change the integration tests to randomly/alternatingly deploy to docker/k8s. This would get us some more coverage without needing to commit more resources.

PR #501, opened 2025-07-10
Fix device connections on node restart

mxsrc:
Restart sporadically fails because upon reconnecting other nodes, controllers might be already present. This PR addresses this by ensuring to reattach the controller upon node restart.

Building up for this, it also includes a few commits introducing helpers to avoid code duplication.


Hamdy-khader (approved):
Looks good, Approved

PR #511, opened 2025-07-16
fix error during sbctl release

boddumanohar:
Currently during sbctl release, make the following change: 

Screenshot 2025-07-16 at 15 11 14

The changes in this PR fixes the need for this commit.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,7 @@ from setuptools import setup, find_packages
 
 from setuptools.command.install import install as _install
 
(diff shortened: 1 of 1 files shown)


mxsrc:
I see. can't we just use a set in that case? If the API requires a list we can convert to list again. That remvoes the duplicate and doesn't need any kind of special handling.

mxsrc (approved):
Looks good! Let's wait for the checks, then it's good to go :)

PR #515, opened 2025-07-16
pin SBCLI & SPDK images for release

boddumanohar:
pin SBCLI & SPDK images for release


mxsrc (approved):
Looks good to me. Did you confirm the workflow file works? I've struggled with the mixture of yaml and bash syntax in the past.

PR #527, opened 2025-07-21
Extending clusters to support being named

noctarius:
This PR adds names to cluster objects. The field is optional and can be set via "--name" or "-n" on cluster create and cluster add operations.

If the name is not explicitly set, or for old cluster objects which were created before the name existed, the cluster listing returns "-" and the API returns null. it should, potentially, be possible to update the cluster name using a new cluster edit operation, but this is out of the scope of this pull request and can be added by a separate one.


noctarius:
@Hamdy-khader can you also have a quick look? You know the FDB part much better :)

PR #530, opened 2025-07-21
improve integration tests run time

boddumanohar:
New run time is: [20m 52s](https://github.com/simplyblock-io/sbcli/actions/runs/16442813975/usage)
Screenshot 2025-07-23 at 12 06 52

## changes

* made `cleanup` stage to run in parallel
* k8s deployment and  creating simplyblock Mgmt cluster can be done in parallel. 

Due to parallelisation of these steps, the run time is reduced from `25 mins` to `20 mins`


mxsrc:
Can you describe the changes you did?

.github/workflows/integration-tests.yml
```diff
@@ -59,36 +59,54 @@ jobs:
...
-    needs: [await-python-checks]
-    runs-on: ubuntu-latest
-    services:
-      docker:
-        image: docker:24.0.5
-        options: --privileged
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v3
+  # cleanup:
```
mxsrc, line 62: Can we either remove this or add a comment explaining why it is kept around?

.github/workflows/integration-tests.yml
```diff
@@ -59,36 +59,54 @@ jobs:
...
+  #                 echo "Removing partition \$part from disk \$disk"
+  #                 sudo parted /dev/\$disk rm \$part || true
+  #               done
+  #             done
+  #           "
+  #         done
 
   integration:
-    needs: [await-image, mock-integration]
+    needs: [await-image, await-python-checks]
```
mxsrc, line 109: It seems you removed the mock-integration. Can we then also remove the `manual_testing/check_endpoints.py` script or is that used elsewhere?

mxsrc (changes requested):
Thanks! Just two minor comments from my side, otherwise it's good to go!

.github/workflows/integration-tests.yml
boddumanohar: removed in [2b87f43](https://github.com/simplyblock-io/sbcli/pull/530/commits/2b87f432f6f508c6d0725ca00b10e089cc6dbcd0)

mxsrc (approved):
Looks good!

PR #541, opened 2025-07-23
Added secret back into the cluster api model

noctarius:
This PR brings back the cluster secret into the public API model (API V2) since it is required for calls further down the API hierarchy.


mxsrc (approved):
Looks good!

PR #565, opened 2025-08-05
Use docker default mode

geoffrey1330:


simplyblock_core/env_var
```diff
@@ -1,5 +1,5 @@
 SIMPLY_BLOCK_COMMAND_NAME=sbcli-dev
-SIMPLY_BLOCK_VERSION=19.2.11
+SIMPLY_BLOCK_VERSION=19.2.12
```
boddumanohar, line 2: Generally, we don't need to this version as everyone has changed to use branch based deployment. Even in SFAM-2270, Alex was using branch based deployment.

boddumanohar (approved):

PR #567, opened 2025-08-05
added second check for enable_unsafe_noiommu_mode

geoffrey1330:


geoffrey1330:
> There is a situation where this does not behave as desired: If the device to bind has an iommu group we should bind it to the vfio-pci driver, irrespective of whether the parameter can be set.

You are suggesting we check if iommu_group exist if tue use vfio-pci?

mxsrc (approved):
Looks good. Let's test whether this works in the deployment, then I think we're good to merge.

PR #606, opened 2025-09-01
add new parameter client_queue_pair_count as a part of cluster create

boddumanohar:
And also as a part of cluster add


Hamdy-khader (approved):
Looks good to me

PR #612, opened 2025-09-02
added support multiple interfaces with prioritized list

geoffrey1330:
We support specifying `ifname` as either a single value or a list. When a list is provided, the interfaces are checked in order of priority.

### DIFF
--- a/simplyblock_core/storage_node_ops.py
+++ b/simplyblock_core/storage_node_ops.py
@@ -817,8 +817,8 @@ def add_node(cluster_id, node_addr, iface_name, data_nics_list,
         if not cloud_instance:
             # Create a static cloud instance from node info
             cloud_instance = {"id": node_info['system_id'], "type": "None", "cloud": "None",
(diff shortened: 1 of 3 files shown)


simplyblock_core/storage_node_ops.py
```diff
@@ -817,8 +817,8 @@ def add_node(cluster_id, node_addr, iface_name, data_nics_list,
         if not cloud_instance:
             # Create a static cloud instance from node info
             cloud_instance = {"id": node_info['system_id'], "type": "None", "cloud": "None",
-                              "ip": node_info['network_interface'][iface_name]["ip"],
-                              "public_ip": node_info['network_interface'][iface_name]["ip"]}
+                              "ip": utils.get_mgmt_ip(node_info, iface_name),
```
boddumanohar, line 820: now that variable `iface_name` takes multiple values. should we also update the variable to represent that its a list?

geoffrey1330: You mean the naming should be changed to **iface_names**

PR #623, opened 2025-09-05
support for rdma (roce)

schmidt-scaled:
this includes support for rdma (roce, but not infiniband!). each NIC can either support tcp or rdma (roce) or both! rdma/roce also uses ip-addresses (ipv4 in our case) and ports. we use exactly the same port numbers as for tcp, but these are udp ports.
when we setup a cluster, we define if it is tcp (default), rdma or both. 
on lvol add we can select the fabric (either tcp or rdma), but the cluster must support the selected fabric, otherwise the lvol add fails.
when we read the list of configured NICs on add-node, we determine the supported protocol (ipv4 and/or roce). we store this information on a per node level. each node must have at least one active interface per supported cluster-fabric-type with an ipv4 address configured. for internal cluster connections, the protocol supported by the cluster AND already active on the target node is used by the client to connect (this is important for migration! some nodes have already both protocols and others only have tcp still, depending if they were restarted already with the new release image). if both protocols are already supported on the target node and the client node, rdma is chosen. 
currently we use a few specific settings for tcp only, but no specific settings for rdma only.


simplyblock_cli/cli-reference.yaml
```diff
@@ -803,6 +803,11 @@ commands:
                 min: 0
                 max: 128
             default: 0
+          - name: "--fabric"
+            help: "fabric: tcp, rdma or both (specify: tcp, rdma)"
+            dest: fabric
+            type: str
```
Hamdy-khader, line 809: Here we can use key "choices" like this:
```
...
type: str
choices:
   - tcp
   - rdma
   - tcp,rdma
default: tcp
...
```

simplyblock_cli/cli-reference.yaml
```diff
@@ -906,6 +911,11 @@ commands:
             dest: enable_node_affinity
             type: bool
             action: store_true
+          - name: "--fabric"
+            help: "fabric: tcp, rdma or both (specify: tcp, rdma)"
+            dest: fabric
+            type: str
```
Hamdy-khader, line 917: Use choices key

simplyblock_core/cluster_ops.py
```diff
@@ -202,12 +202,28 @@ def _set_max_result_window(cluster_ip, max_window=100000):
...
+        "rdma": "rdma" in parts,
+    }
    
 def create_cluster(blk_size, page_size_in_blocks, cli_pass,
                    cap_warn, cap_crit, prov_cap_warn, prov_cap_crit, ifname, mgmt_ip, log_del_interval, metrics_retention_period,
                    contact_point, grafana_endpoint, distr_ndcs, distr_npcs, distr_bs, distr_chunk_bs, ha_type, mode,
                    enable_node_affinity, qpair_count, max_queue_size, inflight_io_threshold, enable_qos, disable_monitoring, strict_node_anti_affinity, name, 
-                   tls_secret, ingress_host_source, dns_name, is_single_node) -> str:
+                   tls_secret, ingress_host_source, dns_name, fabric, is_single_node) -> str:
+                   tls_secret, ingress_host_source, dns_name) -> str:
```
Hamdy-khader, line 226: remove this line
`                   tls_secret, ingress_host_source, dns_name) -> str:`

Hamdy-khader (changes requested):
Looks good, Please fix the inline comments and let us put this branch to test.

simplyblock_cli/cli-reference.yaml
noctarius: Good one. Looks like we may eventually want an oneOrMore rule, but I think for now that's the only one. The choices is a good suggestion though.

PR #642, opened 2025-10-03
Main sfam 2349 (Alceml qos classes)

Hamdy-khader:


schmidt-scaled:
Please have a look at the failing checks. Then we can merge and we will then look at the results of e2e. If it is successful, we are fine to test. If not, we have to look at results of e2e.

Hamdy-khader:
@schmidt-scaled Done, checks are passing now

schmidt-scaled (approved):

PR #643, opened 2025-10-03
Fix SFAM-1465 (pool qos)

Hamdy-khader:
enable pool level qos settings


schmidt-scaled:
Can you please fix the linter? Then we can merge and we will then look at the results of e2e. If it is successful, we are fine to test. If not, we have to look at results of e2e.

Hamdy-khader:
@schmidt-scaled Done, checks are passing now.

PR #669, opened 2025-10-29
Fix port allow hublvol connection

Hamdy-khader:


schmidt-scaled (commented):
the parameter in the check is for the source or the target of the hublvol (from where you connect or to where you connect)?
you need to connect FROM secondary TO primary.

PR #692, opened 2025-11-01
Fix sn suspend set status before port block

Hamdy-khader:
do not check for rule REJECT because it is not sign of an issue


schmidt-scaled (approved):
I will approve it as it is urgent, but I think one thing is needed here still.
If the suspend fails for a reason later, you need to revert the state of suspension to online.

PR #708, opened 2025-11-03
Allow device of state cannot_allocate to start expanding

Hamdy-khader:
Increase the jobs for migrations to 16

Allow all errors in case of a device in state cannot_allocate and read_only in the cluster when migration is completed

### DIFF
--- a/simplyblock_core/services/tasks_runner_migration.py
+++ b/simplyblock_core/services/tasks_runner_migration.py
@@ -2,7 +2,7 @@
 import time
 from datetime import datetime, timezone
 
-from simplyblock_core import db_controller, utils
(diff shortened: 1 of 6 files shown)


simplyblock_core/services/tasks_runner_migration.py
schmidt-scaled: We discussed the rule 1:
If **ANY device is NOT IN (ONLINE, FAILED_AND_MIGRATED)**  and you receive ANY error, you will stop migration on ANY error without immeidate retry.
You said that rule 2 is implemented already:
The migraiton will ALWAYS continue if one devices gets ONLINE again.

PR #720, opened 2025-11-05
do not auto add default qos class

geoffrey1330:


Hamdy-khader (approved):
deploy stage seems to be passing https://github.com/simplyblock/sbcli/actions/runs/19102684746

PR #746, opened 2025-11-17
Update sc name

geoffrey1330:


simplyblock_core/scripts/charts/values-template.yaml
boddumanohar: can you please also remove this.

boddumanohar (approved):

PR #828, opened 2025-12-25
Lvol migration

alexander-sheredin:


simplyblock_core/controllers/lvol_migration_controller.py
```diff
@@ -0,0 +1,638 @@
...
+              elif s.status in ObjectMigrationState.LVOL_CONNECTED:
+                  self.transfer_data(s, 0)
+              elif s.status==ObjectMigrationState.TRANSFERRED:
+                   self.convert_lvol(s,p)
+              elif s.status == ObjectMigrationState.CONVERTED:
+                   self.delete_subsystem(self.m.target_node_pri,s.snap.uuid)
+              elif s.status == ObjectMigrationState.CLEANING:
+                   self.delete_lvol_from_node(self.m.target_node_sec, s.snap.uuid)
+              p=s
+            if self.m.rerun < 3 or self.time_difference()>5:
```
alexander-sheredin: @schmidt-scaled it is not clear where should the counter be incremented (to avoid infinite recursion)

PR #837, opened 2026-01-08
fix jc_suspend_compression usage

Hamdy-khader:


schmidt-scaled (commented):
I am still not sure this is right:
nodes = db_controller.get_primary_storage_nodes_by_secondary_node_id(node.get_id())
                if nodes:
                    sec_node = nodes[0]
                    ret, err = node.rpc_client().jc_suspend_compression(jm_vuid=sec_node.jm_vuid, suspend=False)
                    
you are getting the PRIMARIES by the secondary node id. So in nodes, you have a primary node and thus the sec_node.jm_vuid also contains the PRIMARY lvolstore vuid.

PR #955, opened 2026-04-13
New flag --enable-journal-device

noctarius:
This PR introduces --enable-journal-device and deprecated --journal-partition for better clarity of its usage. Since the values always come down to 0 or 1, a boolean flag disabling default behavior is easier to understand.

### DIFF
--- a/simplyblock_cli/clibase.py
+++ b/simplyblock_cli/clibase.py
@@ -153,7 +153,6 @@ class CLIWrapperBase:
 
         small_bufsize = args.small_bufsize
         large_bufsize = args.large_bufsize
-        num_partitions_per_dev = args.partitions
         jm_percent = args.jm_percent
 
         max_snap = args.max_snap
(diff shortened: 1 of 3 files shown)


simplyblock_cli/clibase.py
```diff
@@ -162,6 +161,19 @@ def storage_node__add_node(self, sub_command, args):
         namespace = args.namespace
         ha_jm_count = args.ha_jm_count
         format_4k = args.format_4k
+
+        if args.enable_journal_device:
+            num_partitions_per_dev = 0
+        else:
+            # Deprecated but still supported for backward compatibility.
+            print("WARNING: --journal-partition is deprecated, use --enable-journal-device instead")
```
mxsrc: The warning prints, irrespective of whether the flag is set explicitly, it should only show when the user actually passed it.

mxsrc (changes requested):
One comment on the deprecation warning, otherwise looks good!

simplyblock_cli/clibase.py
noctarius: You mean, removing the default value and see if it is specifically provided?

noctarius: Good catch!

mxsrc (approved):

PR #975, opened 2026-04-19
Fix misleading --data-nics help text (multiple NICs supported)

navidnadali:
## Summary
- `--data-nics` is defined with `nargs='+'` and `storage_node_ops.add_node` iterates the full list to build per-NIC `IFace` objects (with explicit `len(data_nics) > 1` multipath branches), so the *"Currently, one interface is supported."* help string is wrong.
- The YAML `description` example `eth0,eth1` is actively misleading — argparse splits on whitespace, so a comma-joined value becomes a single bogus token and fails silently downstream (the storage node never registers).
- Updates the help and description to reflect actual behavior; regenerates `cli.py` from `cli-reference.yaml` via `simplyblock_cli/scripts/generate.sh`.

## Test plan
- [ ] `sbctl sn add-node --help` shows the new help text
- [ ] `sbctl sn add-node <cluster> <ip>:5000 eth0 --data-nics eth1 eth2` still registers a node with two data NICs (existing multipath path)
- [ ] No other diffs in `cli.py` from the regeneration

### DIFF
--- a/simplyblock_cli/cli-reference.yaml
+++ b/simplyblock_cli/cli-reference.yaml
@@ -191,9 +191,9 @@ commands:
             default: 3
             private: true
           - name: "--data-nics"
-            help: "The storage network interface names. Currently, one interface is supported."
+            help: "Storage network interface names. Pass one or more space-separated names; multiple NICs enable multipath."
             description: >
-              Storage network interface names, e.g. eth0,eth1
+              Storage network interface names, space-separated, e.g. `eth0 eth1`
             dest: data_nics
             type: str
             nargs: "+"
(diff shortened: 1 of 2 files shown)


noctarius:
I think it's better to fix the comma separated list. It's weird, I'm pretty sure it used to work 🤔
Let me have a look tomorrow.

PS: thanks @navidnadali for flagging it!

noctarius:
Hey @navidnadali, we implemented the actual comma-separated list in #977. Thanks for your help! 🔥

navidnadali:
Thanks @noctarius , Will close this PR

PR #980, opened 2026-04-21
Log API request execution duration

mxsrc:
This disables the uvicorn-builtin access log and adds a middleware instead to log execution duration of each request.


Hamdy-khader (approved):
Thanks!

PR #1003, opened 2026-04-27
Optionally enable TLS

mxsrc:
This PR modifies the clients s.t. they connect via HTTPS instead of plain HTTP mode. This corresponds to the server-side changes of #973. The same heuristic is used to detect the TLS mode: If certificates are present, we are in a TLS deployment. Unless that is the case, the existing logic remains unchanged. In TLS mode, clients now:
- use HTTPS
- verify the presented certificate against the configured CA
- connect to a domain name instead of IPs

The name change is the most notable one, long-term we should target this approach irrespective of transport layer encryption. The operator creates services and endpointslices, s.t. we can connect to:
```
<node>.simplyblock-{storage-node-api,spdk-proxy}.{namespace}.svc.cluster.local
```

It's important that SNodeClient and RPCClient instances are __always__ acquired through the storage node helpers. The old pattern of constructing them and connecting to the management IP will break certificate validation. This needs to be clear to everyone who has been working on sbcli.

The test failures are those present on the main branch.


Hamdy-khader (approved):

geoffrey1330 (approved):
The kubernetes manifests looks good

PR #1018, opened 2026-05-01
snapshot: fix delete race producing stuck open_ref / empty clone entries

schmidt-scaled:
## Summary

Three independent fixes that together close the "Cannot remove snapshot because it is open" / EBUSY (-16) state where the snapshot ends up with non-zero `open_ref` but no clone entries and can only be cleared by restarting the host node (incident: aws_dual_soak 2026-04-30, 14 stuck snapshots).

## Fixes

1. **Random VUID dedupe & range bump** (`simplyblock_core/utils/__init__.py`)
   - `get_random_vuid` range goes from 10k → 1M.
   - Both `get_random_vuid` and `get_random_snapshot_vuid` dedupe against numeric suffixes parsed out of existing `CLN_`/`LVOL_`/`SNAP_` bdev names cluster-wide.
   - With ~10k lvols+snaps the legacy 10k range hit ~50% birthday-collision probability, which is what triggered the SPDK `lvol with name ... already exists` rejection, the mgmt async-delete, and the reuse-during-deletion sequence.

2. **Reject snapshot/clone ops on pending-deletion targets** (`snapshot_controller.add` / `snapshot_controller.clone`)
   - `add` rejects when source lvol is `STATUS_IN_DELETION`.
   - `clone` rejects when source snapshot has `deleted=True` or `STATUS_IN_DELETION`.
   - Closes the window between an async delete being issued and a fresh create slipping through against the same blob.

3. **Snapshot delete waits for clone SPDK delete** (`snapshot_controller.delete`)
   - A clone counts as "gone" only once its `deletion_status` field has been set (i.e. leader's `delete_lvol_from_node` returned). An `IN_DELETION` clone with no `deletion_status` is still SPDK-blocking and the snapshot is soft-deleted instead.
   - The clone's own delete-completion path re-triggers the hard delete once SPDK has actually released the bdev.

## Test plan

- [x] `tests/test_snapshot_delete_race.py` — 10 unit tests covering all three fixes (random vuid dedupe, add/clone rejection, delete waits for in-flight clone)
- [ ] full `tests/` suite green (running)
- [ ] soak verification against AWS cluster after merge

### DIFF
--- a/simplyblock_core/utils/__init__.py
+++ b/simplyblock_core/utils/__init__.py
@@ -358,6 +358,40 @@ def sum_records(records):
         return total
 
 
+_BDEV_NAME_NUMERIC_SUFFIX = re.compile(r'(?:^|[/_])(\d+)\s*$')
+
+
+def _used_bdev_name_numbers(db_controller):
+    """Collect all numeric suffixes already used in lvol/snapshot bdev
+    names cluster-wide (e.g. ``LVS_x/CLN_6900`` -> 6900,
+    ``LVS_x/SNAP_77047`` -> 77047). The clone- and snapshot-create
+    paths build their bdev name as ``CLN_<vuid>``/``SNAP_<vuid>``
+    where ``<vuid>`` comes from the random helpers below; if a fresh
+    random number lands on an already-used suffix SPDK rejects the
+    create with ``lvol with name ... already exists``. The mgmt
+    fallout from that failure is what produced the stuck-snapshot
+    metadata-inconsistency incident (parent's open_ref non-zero,
+    clone entries empty) so we dedupe up-front.
+    """
+    used = set()
+    for lvol in db_controller.get_lvols():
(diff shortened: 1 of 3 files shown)


simplyblock_core/utils/__init__.py
```diff
@@ -358,6 +358,40 @@ def sum_records(records):
         return total
 
 
+_BDEV_NAME_NUMERIC_SUFFIX = re.compile(r'(?:^|[/_])(\d+)\s*$')
+
+
+def _used_bdev_name_numbers(db_controller):
```
Hamdy-khader, line 364: This function can be written like this:

```
def _used_bdev_name_numbers(db_controller):
    used = set()
    for lvol in db_controller.get_lvols():
        used.add(lvol.vuid)

    for snap in db_controller.get_snapshots():
        used.add(snap.vuid)
    return used
```

PR #1058, opened 2026-05-23
exclude IN_DELETION volumes from lvol list and /connect APIs

boddumanohar:


mxsrc (approved):
Good fix, we should really add a helper with a default `include_deleted=False` flag for this. The API needs a query argument for this as well. Neither is urgent though.

noctarius:
We cannot just hide them from the listing. There are tools that need to see them, like the control center. We should have a way to adjust filters. Maybe something like `?status=all` or `?status=online,in_deletion`

noctarius (changes requested):
Please see the comment in slack!

boddumanohar:
I just realised that this could cause inconsistency during volume create with the same name (as the one in_deletion). And an other issue as Chris mentioned that this could also break API contract. So we need to discuss more deeply around this situation needs to be handled given that lvol delete is an asynchronous operation.

PR #1064, opened 2026-05-26
logging: add `requestID` for access logs, application logs and traceback logs

boddumanohar:
currently in the WebApp API we have see 3 types of logs:
1. Access logs --> ex: `POST /cluster/<uid>.. 20ms "Go client`
2. Application logs  --> `ERROR: failed to clone volume`
3. Traceback logs --> `Traceback: ...`

To allow us to connect all these 3 together, have added a new field `request_id` to the logging filter. 

So at the start of the request, within the middleware, we inject the request_id 
```
token = core_utils.request_id_var.set(request_id)
```
and then reset it after the request ends
```
core_utils.request_id_var.reset(token)
```

But since the update has been to the core logger, `simplyblock_core/utils/__init__.py` for CLI, output, we'll see the extra [-] in between.

### DIFF
--- a/simplyblock_cli/clibase.py
+++ b/simplyblock_cli/clibase.py
@@ -6,6 +6,7 @@ import json
 import re
 import sys
 import time
+import uuid
 import argcomplete
 
 from simplyblock_core import cluster_ops, utils, db_controller, constants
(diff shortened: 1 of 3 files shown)


mxsrc (commented):
Generally I like this, but in the case of sbctl execution, instead of `-`, establishing a context var with some random hash might be helpful

mxsrc (approved):

boddumanohar:
rebased the PR. @mxsrc can you please re-review.

mxsrc (changes requested):
I like this, but think the co-pilot suggestions should be folded in. I don't know about the root-logger concern, but if this worked in testing that should be a false-positive. Another thing I'm wondering about is the `X-Request-ID` header. Is that something we'd actually use? Otherwise I'd drop this, not only for the injection concern co-pilot flagged, but just simplicity as well.

PR #1069, opened 2026-05-28
KMS: enforce TLS requirement

boddumanohar:
TLS is hard requirement for KMS feature. So adding a validation as a part of Cluster create.

### DIFF
--- a/simplyblock_core/cluster_ops.py
+++ b/simplyblock_core/cluster_ops.py
@@ -14,6 +14,7 @@ import requests
 
 from docker.errors import DockerException
 from simplyblock_core import utils, scripts, constants, mgmt_node_ops, storage_node_ops
+from simplyblock_core.settings import Settings
 from simplyblock_core.controllers import backup_controller, cluster_events, device_controller, qos_controller, tasks_controller, tcp_ports_events
 from simplyblock_core.fw_api_client import FirewallClient
 from simplyblock_core.db_controller import DBController
@@ -563,6 +564,18 @@ def add_cluster(blk_size, page_size_in_blocks, cap_warn, cap_crit, prov_cap_warn
     cluster.nvmf_base_port = nvmf_base_port
     cluster.rpc_base_port = rpc_base_port
     cluster.snode_api_port = snode_api_port
+    if hashicorp_vault_settings:
+        settings = Settings()
+        missing = [
+            path
+            for path in [settings.tls_certificate_authority, settings.tls_certificate, settings.tls_key]
+            if not path.is_file()
+        ]
+        if missing:
+            raise ValueError(
+                "HashiCorp Vault requires TLS certificates that are not present: "
+                + ", ".join(map(str, missing))
+            )
     cluster.hashicorp_vault_settings = hashicorp_vault_settings
     if backup_config:
         cluster.backup_config = backup_config
--- a/simplyblock_web/api/v2/pool.py
+++ b/simplyblock_web/api/v2/pool.py
@@ -1,7 +1,7 @@
 from typing import Annotated, List, Optional
 from uuid import UUID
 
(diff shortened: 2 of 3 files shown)


simplyblock_web/api/v2/pool.py
```diff
@@ -42,24 +42,23 @@ class StoragePoolParams(BaseModel):
...
         dhchap=parameters.dhchap,
     )
 
-    if not id_or_false:
-        raise ValueError('Failed to create pool')
+    if not pool_id:
+        raise HTTPException(500, 'Failed to create pool')
 
-    pool = db.get_pool_by_id(id_or_false)
-    return pool.to_dict()
```
mxsrc, line 62: Not sure about the modifications here, why are they necessary? In particular:
- The name `id_or_false` is meant to provide some indication of what this value indicates, given the inadequate error reporting of the top-level function
- The other modifications do make some sense, it seems like there was a modification to this endpoint that's inconsistent with the original API design (simply returning an untyped dict instead of the original empty response, I imagine this was a hotfix to avoid a duplicate fetch. A proper way of handling this would be a parameter indicating the desired response).

mxsrc (changes requested):
Thanks for the changes, I have a few comments that merit discussion.

simplyblock_web/api/v2/pool.py
boddumanohar: My implementation is a bit incomplete here. I wanted to have exception based handling rather than depending on `True` or `False`. But then later realised that this will invite a lot of new diff. 


So I'll revert variable change. And then add a query parameter to so that user can decide if they need response or not.

mxsrc: Great, thanks! I don't think we have these parameters in this codebase yet so I think we can just copy the approach from vela. I think this is also an opportunity to properly type the endpoint and return the appropriate DTO instead of a random dict.

simplyblock_core/cluster_ops.py
```diff
@@ -563,6 +564,18 @@ def add_cluster(blk_size, page_size_in_blocks, cap_warn, cap_crit, prov_cap_warn
...
+        missing = [
+            path
+            for path in [settings.tls_certificate_authority, settings.tls_certificate, settings.tls_key]
+            if not path.is_file()
+        ]
+        if missing:
+            raise ValueError(
+                "HashiCorp Vault requires TLS certificates that are not present: "
+                + ", ".join(map(str, missing))
+            )
```
mxsrc, line 578: This was meant to be part of the original review, apparently it got lost in transit:

This makes sense to introduce semantically, but belongs in the settings module imho. I see that `simplyblock_core.settings.Settings.validate_tls_files` fails to extend the check to mTLS:
```
@@ -58,7 +58,7 @@ class Settings(BaseSettings):
         if not self.tls_serve and self.tls_connect == "disabled":
             return self
 
-        if self.tls_serve and (
+        if (self.tls_serve or (self.tls_connect == "authenticated")) and (
             missing := [
                 name
                 for name in ["tls_certificate", "tls_key"]
@@ -66,7 +66,7 @@ class Settings(BaseSettings):
             ]
         ):
             raise ValueError(
-                "SB_TLS_SERVE=true requires TLS files to exist: " + ", ".join(missing)
+                "SB_TLS_SERVE=true/SB_TLS_CONNECT=authenticated require TLS files to exist: " + ", ".join(missing)
             )
 
         if (
```

Once this is fixed, we could simply check `Settings().tls_connect == "authenticated"`, which is much more readable. I can open a PR to introduce the change if you like, given that it's quite small it might make more sense to include in this PR.

boddumanohar: Cool, thanks. would it be great if you can propose changes with a new PR 👍

mxsrc: On it.

PR #1070, opened 2026-05-28
Prevent cross-cluster resource access

mxsrc:
Currently, authentication and resource retrieval is decoupled, allowing API clients to authenticate to one cluster and access resources of a different one. This change introduces checks that resources are accessed in their correct scope.

Note that this fix was complicated by the fact that the backup API does not follow the cannonical API structure. In particular, the deletion of all LVol backups was registered at `/clusters/{cluster_id}/backups/{lvol_id}` instead of `/clusters/{cluster_id}/storage-pools/{pool_id}/volumes/{volume_id}/backups/`. To correct for this, the PR duplicates the endpoint at the latter location and modifies the existing one to call the new function, which implicitly adds the ownership validation. The existing endpoint is marked as deprecated but will continue to function, maintaining compatibility.

I confirmed that the linter and test failures predate this changeset.

Requesting a review from @noctarius specifically to confirm API modification.

### DIFF
--- a/simplyblock_web/api/v2/volume.py
+++ b/simplyblock_web/api/v2/volume.py
@@ -6,12 +6,12 @@ from pydantic import BaseModel, Field, RootModel
 
 from simplyblock_core.db_controller import DBController
 from simplyblock_core import utils as core_utils
-from simplyblock_core.controllers import lvol_controller, snapshot_controller
+from simplyblock_core.controllers import backup_controller, lvol_controller, snapshot_controller
 from simplyblock_core.models.lvol_model import LVol
 
 from .cluster import Cluster
 from .pool import StoragePool
-from .dtos import VolumeDTO, SnapshotDTO, TaskDTO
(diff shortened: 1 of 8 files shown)


Hamdy-khader (approved):
Looks good

noctarius (approved):

simplyblock_web/api/v2/volume.py
```diff
@@ -352,3 +355,24 @@ def clone(
         volume_id=clone_id,
     )
     return Response(status_code=201, headers={'Location': str(entity_url)})
+
+
+@instance_api.get('/backups', name='clusters:storage-pools:volumes:backups:list')
+def backups(volume: Volume) -> List[BackupDTO]:
+    rows = db.get_backups_by_lvol_id(volume.get_id())
```
noctarius, line 362: I guess this would also be a great use case for the indexing.

PR #1077, opened 2026-06-03
hugepages: preserve user's pre-existing reservation per NUMA node

wmousa:
set_hugepages_if_needed now captures the allocatable hugepage count
  for each NUMA node before simplyblock allocates anything, persisting
  it to /tmp/simplyblock/hugepages_baseline_node{N}. On subsequent calls
  within the same boot the cached value is used, so simplyblock's own
  allocations don't inflate the baseline. After reboot /tmp/simplyblock
  is cleared and the baseline is re-sampled from scratch.

  required = user_baseline_per_numa + simplyblock_needed

  Add /tmp/simplyblock hostPath volume to the SNodeAPI Docker container
  and the SPDK pod template so the baseline files live on the host and
  survive container restarts within the same boot.

### DIFF
--- a/simplyblock_core/storage_node_ops.py
+++ b/simplyblock_core/storage_node_ops.py
@@ -2598,6 +2598,7 @@ def _restart_storage_node_impl(
                     snode.ssd_pcie.append(new_ssd)
 
         fdb_connection = cluster.db_connection
+        snode_api.set_hugepages()
         results, err = snode_api.spdk_process_start(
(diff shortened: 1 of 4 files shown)


Hamdy-khader (approved):

boddumanohar:
If we store the file in `/tmp` on the host machine, In Talos, this path is RAM backed tempfs so writing to /tmp doesn't persist node reboot.

boddumanohar:
> live on the host and survive container restarts within the same boot.

Hoo okay, sorry, I got it wrong. So the goal is to survive container reboots. then I think its good.

PR #1085, opened 2026-06-08
Introduce updated authentication/authorization

mxsrc:
This PR updates the webapp API to changes required for a more secure authentication when using the operator. In this mode, the operator and CSI drivers authenticate themselves against the API by providing the token for their service accounts instead of the cluster secret. This is an additional authentication step that may succeed, the existing authentication is untouched.

When deploying, the environment variable `SB_K8S_ADMIN_SERVICE_ACCOUNTS` will be populated with service accounts that are given full authorization to the API. In the future authorization can be expanded s.t. e.g. the CSI will be limited to volume management. Because of this update authorization, the `create-first` cluster creation endpoint, along with its authentication exception could be removed.

Corresponding PRs: https://github.com/simplyblock/simplyblock-operator/pull/233, https://github.com/simplyblock/helm-charts/pull/16, https://github.com/simplyblock/simplyblock-csi/pull/431.

### DIFF
--- a/simplyblock_web/api/v1/cluster.py
+++ b/simplyblock_web/api/v1/cluster.py
@@ -77,68 +77,6 @@ def add_cluster():
     ))
 
 
-@bp.route('/cluster/create_first', methods=['POST'])
-def create_first_cluster():
-    cl_data = request.get_json()
-
-    if db.get_clusters():
-        return utils.get_response_error("Cluster found!", 400)
-
-    blk_size = 512
-    if 'blk_size' in cl_data:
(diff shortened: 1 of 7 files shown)


Hamdy-khader (approved):
Looks good to me

mxsrc:
> Looks good to me

Great. I'll hold the merge until the other PRs are approved as well. The authentication change is safe to introduce on its own, but the `create-first` endpoint is still required with the current operator implementation.

boddumanohar (approved):

PR #1093, opened 2026-06-11
feat: support custom OpenShift MachineConfigPool via OPENSHIFT_MCP en…

### LINKED ISSUE #232: Claim Openshift Custom Machine Config Pool
## Description

OpenShift uses `MachineConfigPool` (MCP) to manage node configurations. When deploying Simplyblock on OpenShift clusters, storage nodes may belong to a custom MCP (e.g. `storage`) rather than the default `worker` pool. The current deployment process does not account for custom MCPs — it assumes all nodes are part of the standard `worker` pool.

This means any `MachineConfig` resources applied during storage node setup (e.g. kernel parameters, hugepages, NVMe modules) target the wrong pool and are never applied to the actual storage nodes, leaving them misconfigured.

## Expected Behaviour

The operator should allow the user to specify a custom MCP name so that `MachineConfig` objects are created with the correct `machineconfiguration.openshift.io/role` label and applied to the right node pool.

## Acceptance Criteria

- A field (e.g. `openShiftMachineConfigPool`) is added to the `StorageNode` CR spec to accept a custom MCP name
- When set, all generated `MachineConfig` resources use the specified pool label instead of `worker`
- When unset, behaviour defaults to `worker` for backward compatibility
- Validated on a cluster with a dedicated storage MCP

geoffrey1330:
…v var in storage node templates

fixes: [issue-232](https://github.com/simplyblock/simplyblock-operator/issues/232)

### DIFF
--- a/simplyblock_web/api/internal/storage_node/kubernetes.py
+++ b/simplyblock_web/api/internal/storage_node/kubernetes.py
@@ -333,6 +333,7 @@ def spdk_process_start(body: SPDKParams):
     if isinstance(skip_kubelet_configuration, str):
        skip_kubelet_configuration = skip_kubelet_configuration.strip().lower() in ("true")
     reserved_system_cpus = os.environ.get("RESERVED_SYSTEM_CPUS", "0,1")
+    openshift_mcp = os.environ.get("OPENSHIFT_MCP")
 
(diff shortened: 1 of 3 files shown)


noctarius:
Looks good but I believe @wmousa should review this.

wmousa (approved):

PR #1094, opened 2026-06-11
Make API creation responses concistent

mxsrc:
Currently, different endpoints for entity creation behave inconsistently:
- Backups: Indicate ID as `X-Backup-Id` header
- Cluster: Return full DTO
- Migration: Return `{migration_id: $id}` object
- Pool: Return database representation
- Storage Node: Return task ID
- Volume: Return empty response with Location header indicating the volume URL

This PR introduces a common helper that supports three approaches, all of which use the `Location` header to indicate the entity location:
- empty response
- ID
- DTO

They can be requested by setting the `response-format` field to `empty`, `identifier`, or `full` respectively. The default for each endpoint ensures this change is backwards compatible with existing clients. In two places, backwards compatibility is not fully given:
- Migration: the object with the `migration_id` is not supported and is replaced by the identifier only response.
- Pools: This returned the database representation which should not be accessible to the user.

The operator is updated to be resilient to the change in return value in https://github.com/simplyblock/simplyblock-operator/pull/241.

Asking for a review from @noctarius as well to gauge the severity of the API stability violation.

### DIFF
--- a/simplyblock_web/api/v2/cluster/__init__.py
+++ b/simplyblock_web/api/v2/cluster/__init__.py
@@ -1,6 +1,8 @@
 from threading import Thread
 from typing import Annotated, List, Literal, Optional
-from fastapi import APIRouter, HTTPException, Response
+from uuid import UUID
+
+from fastapi import APIRouter, HTTPException, Request, Response
 from pydantic import BaseModel, Field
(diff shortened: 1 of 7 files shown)


Hamdy-khader (approved):
Looks good

PR #1096, opened 2026-06-12
Fix volume resize on equal size

mxsrc:
Currently, resizing a volume to its current size fails with an error. While this is technically correct --- the best kind of correct --- it's unexpected for users. This PR introduces an early abort in this case, while reporting success. I used the opportunity to rework the `lvol_controller.resize_lvol` function to exception-based error handling.

### DIFF
--- a/simplyblock_cli/clibase.py
+++ b/simplyblock_cli/clibase.py
@@ -613,8 +613,8 @@ class CLIWrapperBase:
     def volume__resize(self, sub_command, args):
         volume_id = args.volume_id
         size = args.size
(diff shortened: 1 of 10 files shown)


noctarius:
Looks good to me.

FYI @boddumanohar

Hamdy-khader (approved):

mxsrc:
I added one more commit to avoid human friendly formatting of sizes in error messages when the size exceeds limits. The actual difference might get lost in rounding otherwise, e.g. "New size 1024.0MiB must not be lower than current size 1GiB"

noctarius (approved):

PR #1097, opened 2026-06-12
Single-node cluster expansion (FTT1/FTT2)

schmidt-scaled:
## Summary

Adds single-node cluster expansion: integrate a newcomer storage node and rotate sec/tert LVS roles across the cluster, reusing the existing restart/activate primitives.

### Feature
- `--expansion` CLI flag + donor-side helpers; orchestrator triggers expansion migration after `integrate_new_node_into_cluster`.
- Planner derives role moves from host topology (`compute_role_diff_topology`, `_host_rotation_layout`, `_validate_topology`); sec/tert slot naming aligned to `lvstore_stack_secondary/tertiary` + `tertiary_node_id`.
- Executor (`SpdkMoveExecutor`) drives create-primary / create-sec/-tert / re-home moves; donor cleanup via `teardown_non_leader_lvstore`.
- Offline protocol-simulation harness: expands the planner's move list into the per-move RPC sequence and asserts structural invariants, emitting reviewable RPC-by-RPC traces — no FoundationDB/executor required.

### Shared-path integration & hardening
Expansion reuses `recreate_lvstore_on_non_leader` / `create_lvstore` (called with `activation_mode=True`), so a few changes land on the shared restart/activate path:
- `recreate_lvstore_on_non_leader`: per-LVS `lvstore_ports` rebuild + early `connect_to_hublvol` (the only hublvol connect that runs under `activation_mode`).
- `create_lvstore`: always create the primary hublvol (newcomer gains secondaries post-creation via rotation).
- **Hardening:** None-guard the `hublvol.nvmf_port` derefs in the port rebuild (legacy/single-node nodes with no hublvol previously raised `AttributeError` — soft-failing secondary recreation on restart, hard-failing cluster activate). Removed a dead duplicate `elif STATUS_DOWN` branch in the monitor.

### Known residual risk (reviewers please note)
- `lvstore_ports` source-of-truth changed (rebuilt from node's own ports + back-refs vs copied from primary). Expected equivalent for single-secondary restart; worth a restart/activate smoke test.
- `connect_to_hublvol` fires unconditionally during `activation_mode`; harmless (try/except) during activate but premature — clean fix would be a dedicated param to separate activate-defer vs expansion-connect-now.

## Test
- Local `pytest` suite running; CI runs the full suite + lint + cli-gen check against this PR.


Hamdy-khader (approved):
Looks good, approved.

simplyblock_core/env_var
```diff
@@ -1,5 +1,5 @@
 SIMPLY_BLOCK_COMMAND_NAME=sbcli-dev
 SIMPLY_BLOCK_VERSION=19.2.34
 
-SIMPLY_BLOCK_DOCKER_IMAGE=simplyblock/simplyblock:main
-SIMPLY_BLOCK_SPDK_ULTRA_IMAGE=simplyblock/spdk:main-latest
+SIMPLY_BLOCK_DOCKER_IMAGE=public.ecr.aws/simply-block/simplyblock:feature-single-node-expansion
+SIMPLY_BLOCK_SPDK_ULTRA_IMAGE=public.ecr.aws/simply-block/ultra:main-latest
```
mxsrc: This should probably not be part of this PR.

single_node_expansion_plan.md
```diff
@@ -0,0 +1,177 @@
+# Single-node cluster expansion (FTT2) — implementation plan
```
mxsrc: This file should not be added to the repo. I think it may make sense to document the changes, but that should be placed in `docs/`, and not take the form of a planning document.

step6_design.md
```diff
@@ -0,0 +1,163 @@
+# Step 6 design — final
```
mxsrc: This should not be introduced to the repository.

simplyblock_core/rpc_client.py
```diff
@@ -379,6 +379,19 @@ def bdev_nvme_detach_controller(self, name):
...
+        detaching the whole controller. Used during single-node cluster
+        expansion to prune the dead failover path on a sibling sec_2 node
+        after its sec_1 was re-homed to a different node."""
+        params = {
+            "name": name,
+            "traddr": traddr,
+            "trsvcid": str(trsvcid),
+            "trtype": trtype,
+        }
+        return self._request2("bdev_nvme_remove_trid", params)
```
mxsrc: New functions should utilize the `_request3` helper which uses exception-based error handling.

mxsrc (changes requested):
I couldn't do a thorough review yet, but superficially there are a few things that should be changed before merging this. In addition to the comments on specific files the tests that are introduced do not succeed and their execution takes >2h, making them prohibitively expensive for any reasonable checks during development. The long execution time may be a result of the failure, but if it's not, we should introduce some sort of multi-step testing to make test driven development feasible, relegating long running tests to the last step. The failing unittests are probably a false-positive, I assume the general feature works as intended, still, this should be addressed before integration.

The actual logic for the feature seems to not use the established task mechanism that other asynchronous tasks use, like storage node addition. Is that a deliberate choice? Generally, I think the code should be placed in a shared module, e.g. `simplyblock_core/controllers/cluster_expansion/...` instead of at the top-level of `simplyblock_core`. I'll do a full review of the logic tomorrow, right now this is just a high-level comment.

simplyblock_core/cluster_expand_orchestrator.py
```diff
@@ -0,0 +1,190 @@
...
+    def execute(self, move: RoleMove) -> None:
+        self.executed.append(move)
+
+
+class ExpandPlanError(Exception):
+    """Raised by ``execute_expand_plan`` when a fatal precondition is not
+    met (e.g., trying to start a fresh plan while one is already in
+    progress, or resuming an incompatible-schema state). Distinct from the
+    underlying executor exception that triggers an abort, which is
+    re-raised as-is after the abort is persisted."""
```
mxsrc: There is a generic `PreconditionError` in `simplyblock_core/exceptions` that fits this purpoes.

PR #1098, opened 2026-06-12
Lvol migration fresh

EbiRider:
this aims to add the the migration feature to the sbcli and web api 
this includes a few new calls 
migration list
migrate 
pre-create-migration
migrate-cancel


simplyblock_cli/cli-reference.yaml
```diff
@@ -1922,12 +1922,12 @@ commands:
         help: "Migrate a logical volume to a different storage node."
         private: true
         arguments:
-          - name: "volume_id"
-            help: "The logical volume id."
-            dest: volume_id
+          - name: "volume_or_migration_id"
```
mxsrc: Any name that includes `or` or `and` indicates something is amiss. In this case, we should add two individual arguments for both options and ensure that either is set.

simplyblock_core/controllers/lvol_controller.py
```diff
@@ -2450,12 +2448,20 @@ def replication_start(lvol_id, replication_cluster_id=None):
...
-                task = tasks_controller.add_snapshot_replication_task(snap.cluster_id, snap.lvol.node_id, snap.get_id())
-                if task:
-                    snapshot_events.replication_task_created(snap)
+                for sn in all_snaps:
+                    if sn.lvol.node_id == lvol.replication_node_id and sn.data_uuid == snap.data_uuid:
+                        snap = db_controller.get_snapshot_by_id(snap.get_id())
+                        snap.target_replicated_snap_uuid = sn.get_id()
+                        snap.write_to_db()
+                        break
+                else:
```
mxsrc: We should not use the `for ... else` construct, no one knows the exact syntax.

simplyblock_core/controllers/lvol_controller.py
```diff
@@ -2450,12 +2448,20 @@ def replication_start(lvol_id, replication_cluster_id=None):
...
+                    task = tasks_controller.add_snapshot_replication_task(snap.cluster_id, snap.lvol.node_id, snap.get_id())
+                    if task:
```
mxsrc, line 2343: The if statement here implies the function can not return a task. That is probably an error which is silently ignored here. Is that intentional? If so, it deserves a comment that justifies it.

simplyblock_core/controllers/migration_controller.py
```diff
@@ -59,22 +63,30 @@
...
-    - target node exists, is online, and is not the current node
-    - no active migration already running for this volume on its source node
-    - cluster is in an active state
+    Returns (migration_uuid, None) on success or (False, error_message) on failure.
     """
-    # --- Validate volume ---
+    try:
+        migration = db.get_migration_by_id(migration_id)
+    except KeyError:
+        return False, f"Migration {migration_id} not found"
```
mxsrc: The migration feature is newly introduced, as such it should use exceptions, not return codes for error management.

simplyblock_core/controllers/migration_controller.py
```diff
@@ -59,22 +63,30 @@
...
+
+    if migration.phase != LVolMigration.PHASE_PRE_CREATED:
+        return False, (
+            f"Migration {migration_id} is not in PHASE_PRE_CREATED "
+            f"(phase={migration.phase})"
+        )
+
+    lvol_id        = migration.lvol_id
```
mxsrc: Assignments should not be aligned with whitespaces in Python.

simplyblock_core/constants.py
```diff
@@ -319,7 +287,8 @@ def get_config_var(name, default=None):
 # Live volume migration constants
 LVOL_MIG_MAX_RETRIES = 5          # max retry attempts before aborting
 LVOL_MIG_DEADLINE_SEC = 360  # default 4-hour deadline (0 = no deadline)
-LVOL_MIG_MAX_INTERMEDIATE_SNAPS = 3  # max recursive "shrink" snapshot rounds
+LVOL_MIG_MAX_INTERMEDIATE_SNAPS = 3        # max recursive "shrink" snapshot rounds
+LVOL_MIG_INTERMEDIATE_SNAP_THRESHOLD_BYTES = 500 * 1024 * 1024  # 500 MB — skip if delta is smaller
```
mxsrc: This unit conversion is wrong, it's not MB but MiB.

Do not use manual computation for unit conversion, utilize the helpers in simplyblock_core.utils.

simplyblock_core/env_var
```diff
@@ -1,5 +1,5 @@
 SIMPLY_BLOCK_COMMAND_NAME=sbcli-dev
 SIMPLY_BLOCK_VERSION=19.2.34
 
-SIMPLY_BLOCK_DOCKER_IMAGE=simplyblock/simplyblock:main
-SIMPLY_BLOCK_SPDK_ULTRA_IMAGE=simplyblock/spdk:main-latest
+SIMPLY_BLOCK_DOCKER_IMAGE=public.ecr.aws/simply-block/simplyblock:lvol-migration-fresh
+SIMPLY_BLOCK_SPDK_ULTRA_IMAGE=public.ecr.aws/simply-block/ultra:lvol-migration2-slim
```
mxsrc, line 5: This should not be merged to main.

simplyblock_core/controllers/migration_controller.py
```diff
@@ -468,28 +651,474 @@ def apply_migration_to_db(migration):
...
+        src_node = db.get_storage_node_by_id(src_node_id)
+    except KeyError:
+        raise ValueError(f"Source node {src_node_id} not found")
+
+    cluster = db.get_cluster_by_id(tgt_node.cluster_id)
+    tgt_rpc  = tgt_node.rpc_client()
+    nqn      = lvol.nqn
+    bdev_short  = lvol.lvol_bdev + _MIGRATION_BDEV_SUFFIX
+    composite   = f"{tgt_node.lvstore}/{bdev_short}"
+    size_in_mib = lvol.size // (1024 * 1024)
```
mxsrc: Do not use manual unit conversion, utilize the helpers in `simplyblock_core.utils`.

simplyblock_core/services/tasks_runner_lvol_migration.py
```diff
@@ -763,7 +905,25 @@ def _handle_snap_copy(migration, src_node, tgt_node, src_rpc, tgt_rpc):
...
+        if _delta is not None and _delta <= _threshold:
+            logger.info(
+                f"Intermediate snapshot skipped: delta {_delta // (1024 * 1024)} MB "
+                f"<= {_threshold // (1024 * 1024)} MB threshold "
+                f"(round {migration.intermediate_snap_rounds}/{migration.max_intermediate_snap_rounds})")
+            break
+        logger.info(
+            f"Intermediate snapshot triggered: delta "
+            f"{'unknown' if _delta is None else str(_delta // (1024 * 1024)) + ' MB'} "
+            f"exceeds {_threshold // (1024 * 1024)} MB threshold")
```
mxsrc: Do not use manual unit conversion, utilize the helpers in `simplyblock_core.utils`.

simplyblock_core/controllers/migration_controller.py
```diff
@@ -105,34 +116,30 @@ def start_migration(lvol_id, target_node_id,
...
+                    instance_found = True
+                    break
+        if instance_found:
+            snaps_found_on_target.append(snap_id)
+
+    snap_migration_plan = []
+    for snap_id in snap_plan:
+        if snap_id not in snaps_found_on_target:
+            snap_migration_plan.append(snap_id)
+
```
mxsrc: Code like this is difficult to parse. Introduce and use helpers for this:
```
snapshot_migration_plan = [snapshot for snapshot in get_snapshot_chain(lvol_id, source_node_id) if snapshot.preset_on(target_node_id)]
```

simplyblock_core/controllers/migration_controller.py
```diff
@@ -105,34 +116,30 @@ def start_migration(lvol_id, target_node_id,
...
     migration.source_node_id = source_node_id
-    migration.target_node_id = target_node_id
     migration.phase = LVolMigration.PHASE_SNAP_COPY
-    migration.snap_migration_plan = snap_plan
+    migration.snap_migration_plan = snap_migration_plan
```
mxsrc, line 138: This should not be pre-computed, it needs be part of the task execution. The presence of snapshots may also be modified by outside factors, so the task needs to be resilient to this. The task needs to ensure all snapshots that are required are present on the target node, and it needs to use a mechanism to lock snapshots from deletion.

mxsrc (changes requested):
Please provide a description for future PRs that elaborates on the motivation for the PR, the design of the changes, and any particularly noteworthy locations that are difficult to parse.

I have a number of high and low-level comments. From a interface design perspective I believe the most pressing issue is that the `pre_create` mechanic is not a step of the migration task. This is unintuitive. In fact, the `pre-create` function returns a `migration_id`, so it _does_ create the migration, it should be the first step during migration execution. This should also be reflected in the API, instead of:
```
POST .../migrations/pre-create
POST .../migrations/{migration_id}/
```
it should be
```
POST .../migrations/
POST .../migrations/{migration_id}/continue
```
likewise for the CLI interface. It may even be feasible to remove the manual instruction to advance the migration, neither the PR nor the code documentation provides any information on the motivation for this design choice. As it is, it feels like it's unnecessary.
EDIT: Chris explained to me why the manual intervention is required, still this is clear from neither the PR nor the code. Also the point about the API design stands.

Another issue that is pre-existing but should be addressed in this PR is that the migrations API is lvol specific. The endpoints should be at `/clusters/{cluster_id}/storage-pools/{pool_id}/volumes/{volume_id}/migrations/`
instead of `/clusters/{cluster_id}/migrations/`, with the corresponding changes to the parameters because the volume can now be derived from the path.

Other general comments that I added in some places, but got too lazy to do it consistently:
- use helpers. Whenever you add a comment about what a block of code does --- that should be a function. When you introduce 20 helpers for a function, put that function and its helpers into a dedicated submodule
- do not align assignments with whitespaces: https://peps.python.org/pep-0008/#whitespace-in-expressions-and-statements
- do not manually convert units, we have helpers for that which avoid introducing confusing decimal and binary units.

simplyblock_core/controllers/migration_controller.py
```diff
@@ -431,13 +555,25 @@ def _collect_snap_ancestry(snap_uuid, out_set):
...
 
-def apply_migration_to_db(migration):
+# Suffix appended to every bdev the task runner creates on the target node
+# during migration. Prevents accidental collision with real pre-existing target
+# bdevs on retry / initial attempts. Must match _MIGRATION_BDEV_SUFFIX in the
+# task runner.
+_MIGRATION_BDEV_SUFFIX = 'm'
+
+
+def apply_migration_to_db(migration, tgt_lvol_uuid=None, tgt_lvol_bdev=None):
```
mxsrc: This should be a step in the migration itself, placing it here makes the command flow unintuitive.

simplyblock_core/controllers/migration_controller.py
```diff
@@ -454,9 +590,56 @@ def apply_migration_to_db(migration):
         logger.error(f"apply_migration_to_db: target node not found: {e}")
         return False
 
-    lvol.node_id = tgt_node.get_id()
+    # --- Query SPDK once for all bdevs on the target lvstore ---
```
mxsrc: Whereever we feel it's necessary to use a comment to indicate what the next part of the code does, a function should be created for this purpose.

This function has a cyclomatic complexity of 39, the limit linters commonly recommend is 15.

simplyblock_core/controllers/migration_controller.py
```diff
@@ -468,28 +651,474 @@ def apply_migration_to_db(migration):
...
+            else:
+                # If the snapshot belongs to this LVol, then it is fine to transfer it to the target node
+                snap.write_to_db(db.kv_store)
+
+        logger.debug(f"apply_migration_to_db: updated snapshot {snap_uuid} snap_bdev={snap.snap_bdev}")
 
     return True
 
 
+def pre_create_on_target(lvol_id, target_node_id,
```
mxsrc: This function has a cyclomatic complexity of 74. Please use helpers for individual components. This is probably large enough to introduce a `migration_controller` submodule.

simplyblock_core/controllers/migration_controller.py
```diff
@@ -468,28 +651,474 @@ def apply_migration_to_db(migration):
...
+    except KeyError:
+        raise ValueError(f"Source node {src_node_id} not found")
+
+    cluster = db.get_cluster_by_id(tgt_node.cluster_id)
+    tgt_rpc  = tgt_node.rpc_client()
+    nqn      = lvol.nqn
+    bdev_short  = lvol.lvol_bdev + _MIGRATION_BDEV_SUFFIX
+    composite   = f"{tgt_node.lvstore}/{bdev_short}"
+    size_in_mib = lvol.size // (1024 * 1024)
+    tgt_port    = tgt_node.get_lvol_subsys_port(tgt_node.lvstore)
```
mxsrc: Assignments should not use whitespaces for alignment: https://peps.python.org/pep-0008/#whitespace-in-expressions-and-statements

simplyblock_core/controllers/migration_controller.py
```diff
@@ -468,28 +651,474 @@ def apply_migration_to_db(migration):
...
+        except Exception as _e:
+            logger.warning(
+                f"pre_create_on_target: secondary registration error (continuing): {_e}")
+
+    # ── 2. Subsystem + listeners on all TGT nodes ────────────────────────────
+    # Compute overlap: node IDs present in both SRC path and TGT path.
+    # Overlap nodes already own a live subsystem from their SRC role — we add
+    # only an inaccessible listener at the TGT port.  Non-overlap nodes get a
+    # full subsystem, listeners, allowed_hosts, and a namespace pointing to the
+    # migration bdev so clients can pre-connect before cutover.
```
mxsrc, line 770: Not sure whether that's feasible here, but generally for this kind of thing it's great to use idempotent helpers of the `ensure_X` sort. They can be called to ensure the condition is met, and will return early if this is already the case. They also aid in reducing the exposure to timing issues between the check and instantiation.

simplyblock_core/controllers/migration_controller.py
```diff
@@ -468,28 +651,474 @@ def apply_migration_to_db(migration):
...
+                "ctrl-loss-tmo": ctrl_loss_tmo,
+                "fast_io_fail_tmo": constants.LVOL_NVME_CONNECT_FAST_IO_FAIL_TO,
+                "nr-io-queues": cluster.client_qpair_count,
+                "keep-alive-tmo": keep_alive_tmo,
+                "host-iface": cluster.client_data_nic,
+                "connect": connect_cmd,
+            }
+            if host_entry and host_entry.get("psk"):
+                entry["tls"] = True
+            out.append(entry)
```
mxsrc: This duplicates the code in `lvol_controller.connect_lvol`, there must be some common components that can be extracted to a helper.

simplyblock_core/env_var
EbiRider: yes I'm aware actually i don't know why the env_var is not git ignored each time i rebase there's a commit that changes it, could have a env_var.example if we want to keep an updated file for reference but i guess it's required for the build image since i had to push a commit to get correct builds with relevant tags , anyway i have a note to change the env_var before merging this PR

simplyblock_core/controllers/health_controller.py
```diff
@@ -988,8 +968,10 @@ def check_lvol_on_node(lvol_id, node_id, node_bdev_names=None, node_lvols_nqns=N
             bdev_name = bdev_info['name']
             if bdev_info['type'] in ["bdev_lvol", "bdev_lvol_clone"]:
                 bdev_name = lvol.lvol_uuid
-
-            passed &= check_bdev(bdev_name, bdev_names=node_bdev_names)
+            bdev_check = check_bdev(bdev_name, bdev_names=node_bdev_names)
+            if not bdev_check:
+                bdev_check = check_bdev(lvol.top_bdev, bdev_names=node_bdev_names)
+            passed &= bdev_check
```
mxsrc, line 1000: What about using this instead?
```
passed &= check_bdev(bdev_name, bdev_names=node_bdev_names) or check_bdev(lvol.top_bdev, bdev_names=node_bdev_names)
```
Also I don't understand why that's required here. Might make sense to introduce a short comment explaining why checking the bdev name is expected to fail, and why `top_bdev` is a fallback in that case? It looks like we'll be repeating the check for the top_bdev for any bdev check failure, is that intended?

simplyblock_core/controllers/migration_controller.py
```diff
@@ -59,80 +65,68 @@
...
-    cluster_id = source_node.cluster_id
-
-    # --- Check for conflicting active migration on the same source node ---
-    existing = get_active_migration_on_node(cluster_id, source_node_id)
-    if existing:
-        return False, (
-            f"Another migration is already active on source node {source_node_id} "
-            f"(migration_id={existing.uuid})"
-        )
+        raise ValueError(f"Target node is not online (status={target_node.status})")
```
mxsrc, line 123: This can be simplified significantly:
```
    try:
        migration = db.get_migration_by_id(migration_id)

        if migration.phase != LVolMigration.PHASE_PRE_CREATED:
            raise PreconditionError(
                f"Migration {migration_id} is not in PHASE_PRE_CREATED "
                f"(phase={migration.phase})"
            )

        lvol = db.get_lvol_by_id(migration.lvol_id)

        if lvol.status != LVol.STATUS_ONLINE:
            raise PreconditionError(f"Volume is not online (status={lvol.status})")

        source_node = db.get_storage_node_by_id(lvol.node_id)
        target_node = db.get_storage_node_by_id(migration.target_node_id)

        if source_node.get_id() == target_node.get_id():
            raise PreconditionError("Source and target nodes must be different")

        if source_node.status != StorageNode.STATUS_ONLINE:
            raise PreconditionError(f"Source node is not online (status={source_node.status})")

        if target_node.status != StorageNode.STATUS_ONLINE:
            raise PreconditionError(f"Target node is not online (status={target_node.status})")
    except KeyError as e:
        raise PreconditionError("Required entity missing") from e
```

A few comments on this:
- when raising an error based on a prior error, the idiom is to do this to clearly, consistently indicate the fault chain and implicitly copy the error message
try:
    operation()
except A as e:
    raise B("operation failed") from e
- this handling joins the try blocks, making it much easier to parse that the block checks preconditions cyclomatic complexity drops from 15 to 9 for the whole function.
- Uses the recently introduced `PreconditionError` to distinguish a failure of the operation from unmet preconditions (sorry for missing this in the previous review)
- Removes unnecessary variables for IDs: instead of storing `entity_id` and `entity` separately, we only store the `entity` and retrieve the ID using `.get_id()` later. This introduces a single source of truth and avoids both coming out of sync in any way.

simplyblock_core/controllers/migration_controller.py
```diff
@@ -141,45 +135,158 @@ def start_migration(lvol_id, target_node_id,
...
-    return True, None
+
+
+def _cleanup_pre_created(migration):
+    """
+    Delete the subsystem(s) and migration bdev that were created during
+    pre_create_on_target.  Called inline by cancel_migration when
+    phase == PHASE_PRE_CREATED (no task runner is involved yet).
+
+    Best-effort: logs errors but does not raise.
```
mxsrc, line 212: Is this appropriately handled in the migration itself? What happens if I cancel a migration, any of the cleanup components errors, and I try again? Will the pre-existing state cause issues?

simplyblock_core/controllers/migration_controller.py
```diff
@@ -141,45 +135,158 @@ def start_migration(lvol_id, target_node_id,
...
         migration = db.get_migration_by_id(migration_id)
     except KeyError as e:
-        return False, str(e)
+        raise ValueError(str(e))
 
     if not migration.is_active():
-        return False, f"Migration is not active (status={migration.status})"
+        raise ValueError(f"Migration is not active (status={migration.status})")
+
+    if migration.phase == LVolMigration.PHASE_PRE_CREATED:
```
mxsrc, line 192: This reinforces my above comment about folding the `pre-create` step into regular execution. It would avoid having to make a distinction here and keep things self-contained in the task execution.

simplyblock_core/controllers/migration_controller.py
```diff
@@ -141,45 +135,158 @@ def start_migration(lvol_id, target_node_id,
...
+        lvol = db.get_lvol_by_id(migration.lvol_id)
+    except KeyError:
+        logger.warning(f"_cleanup_pre_created: lvol {migration.lvol_id} not found — skipping cleanup")
+        return
+
+    try:
+        tgt_node = db.get_storage_node_by_id(migration.target_node_id)
+    except KeyError:
+        logger.warning(f"_cleanup_pre_created: target node {migration.target_node_id} not found — skipping cleanup")
+        return
```
mxsrc, line 224: I suggest the same change about exception handling as above, fold blocks into one, raise PreconditionError.

simplyblock_core/controllers/migration_controller.py
```diff
@@ -214,8 +321,11 @@ def is_migration_active_on_node(node_id, cluster_id=None):
 def list_migrations(cluster_id=None, is_json=False):
     """Return a formatted list (table or JSON) of all migrations."""
     migrations = db.get_migrations(cluster_id)
+    migrations = sorted(migrations, key=lambda x: x.create_dt)
+
     data = []
     for m in reversed(migrations):  # newest first
+        logger.debug(m)
```
mxsrc, line 339: This seems like a mistake, if we want to get all migrations for debug purposes, we can query the database. Dumping all migrations on a query endpoint is not appropriate.

simplyblock_core/controllers/migration_controller.py
```diff
@@ -214,8 +321,11 @@ def is_migration_active_on_node(node_id, cluster_id=None):
...
         data.append({
```
mxsrc, line 340: I know this isn't introduced here, but this should return the `MigrationDTO`. Might make sense to adapt if this part is touched in the code anyway.

simplyblock_core/controllers/migration_controller.py
```diff
@@ -297,6 +407,18 @@ def _add(uid):
     return result
 
 
+def _is_snap_on_node(snap_id, node_id):
+    """Return True if *snap_id* already has an instance on *node_id*."""
+    try:
+        snap = db.get_snapshot_by_id(snap_id)
```
mxsrc, line 424: This introduces lots of duplicated lookups. Why not pass the entity?

simplyblock_core/controllers/migration_controller.py
```diff
@@ -297,6 +407,18 @@ def _add(uid):
...
+    except KeyError:
+        return False
```
mxsrc, line 426: This is not exception-based error handling.

simplyblock_core/controllers/migration_controller.py
```diff
@@ -431,65 +553,336 @@ def _collect_snap_ancestry(snap_uuid, out_set):
 # Post-migration DB updates
 # ---------------------------------------------------------------------------
 
-def apply_migration_to_db(migration):
+_MIGRATION_BDEV_SUFFIX = constants.LVOL_MIG_BDEV_SUFFIX
```
mxsrc, line 567: This introduces a level of indirection for no reason. Why not just use `constants.LVOL_MIG_BDEV_SUFFIX` at the usage-site?

simplyblock_core/controllers/migration_controller.py
```diff
@@ -431,65 +553,336 @@ def _collect_snap_ancestry(snap_uuid, out_set):
...
+            "fast_io_fail_tmo": constants.LVOL_NVME_CONNECT_FAST_IO_FAIL_TO,
+            "nr-io-queues": cluster.client_qpair_count,
+            "keep-alive-tmo": keep_alive_tmo,
+            "host-iface": cluster.client_data_nic,
+            "connect": connect_cmd,
+        }
+        if host_entry and host_entry.get("psk"):
+            entry["tls"] = True
+        entries.append(entry)
+    return entries
```
mxsrc, line 616: This is still the same code as the one in `lvol_controller.connect_lvol`. You moved the code to a helper (good), but still introduced duplication (bad). There should be a reusable helper for this task that's used in both the `migration_controller` and the `lvol_controller`.

simplyblock_core/controllers/migration_controller.py
```diff
@@ -431,65 +553,336 @@ def _collect_snap_ancestry(snap_uuid, out_set):
...
-    lvol.nodes = [tgt_node.get_id()]
-    if tgt_node.secondary_node_id:
-        lvol.nodes.append(tgt_node.secondary_node_id)
-    if tgt_node.tertiary_node_id:
-        lvol.nodes.append(tgt_node.tertiary_node_id)
+    src_node_id = lvol.node_id
+    try:
+        src_node = db.get_storage_node_by_id(src_node_id)
+    except KeyError:
+        raise ValueError(f"Source node {src_node_id} not found")
```
mxsrc, line 669: Same comment as on `migration_controller.start_migration` applies.

simplyblock_core/controllers/migration_controller.py
mxsrc: This still applies, the complexity is a bit lower but still too high, at 54. The comment markers are great starters to extract sub-functions for the individual tasks.

simplyblock_core/controllers/snapshot_controller.py
```diff
@@ -351,20 +341,39 @@ def add(lvol_id, snapshot_name, backup=False, lock=True, all_snaps=None, all_lvo
     return snap.uuid, False
 
 
-def list(all=False, cluster_id=None, with_details=False, pool_id_or_name=None):
-    if pool_id_or_name:
+def list_snapshots(cluster_id=None, node_id=None, lvol_id=None,pool_id_or_name=None, with_details=False, is_json=False):
```
mxsrc, line 369: This makes the function unnecessarily complex:
- `is_json` does not belong at this level of abstraction, the caller can decide to format the data in json if necessary, and
- `id_or_name` should be resolved before this function.

I suggest introducing
```
`list_snapshots` to `lvol_controller`, `pool_controller`, and `storage_node_ops` that take an _entity_ of the the correct type each:
```
pool_controller.list_snapshots(db.get_pool_by_name(pool_name))
```

File: simplyblock_core/services/tasks_runner_lvol_migration.py [hunk 31]
```diff
@@ -128,22 +134,318 @@ def _get_migration_nic(node):
...
+            )
+            src_rpc.bdev_nvme_detach_controller(ctrl_name)
+            ret = src_rpc.bdev_nvme_attach_controller(
+                ctrl_name, mig_hub_nqn, target_ip, hub_port, trtype)
+        if not ret:
+            return None, None, f"Failed to attach migration hub controller to {tgt_node.get_id()}"
+    return ctrl_name, hub_bdev, None
+
+
+_MIGRATION_BDEV_SUFFIX = constants.LVOL_MIG_BDEV_SUFFIX
```
mxsrc, line 201: See above.

simplyblock_core/services/tasks_runner_lvol_migration.py
```diff
@@ -128,22 +134,318 @@ def _get_migration_nic(node):
...
+            logger.info(
+                f"_ensure_hub_attached: zombie mighub controller found (no bdev); "
+                f"detaching and reattaching"
+            )
+            src_rpc.bdev_nvme_detach_controller(ctrl_name)
+            ret = src_rpc.bdev_nvme_attach_controller(
+                ctrl_name, mig_hub_nqn, target_ip, hub_port, trtype)
+        if not ret:
+            return None, None, f"Failed to attach migration hub controller to {tgt_node.get_id()}"
+    return ctrl_name, hub_bdev, None
```
mxsrc: Is there any common ground with the existing `create_hublvol` helpers in `simplyblock_core.models.storage_node.StorageNode` that might be shared?

simplyblock_core/services/tasks_runner_lvol_migration.py
```diff
@@ -128,22 +134,318 @@ def _get_migration_nic(node):
...
+        lvol = db.get_lvol_by_id(migration.lvol_id)
+    except KeyError as e:
+        logger.error(f"_apply_migration_to_db: lvol not found: {e}")
+        return False
+
+    try:
+        tgt_node = db.get_storage_node_by_id(migration.target_node_id)
+    except KeyError as e:
+        logger.error(f"_apply_migration_to_db: target node not found: {e}")
+        return False
```
mxsrc, line 230: Same as above, also not using exceptions.

simplyblock_core/rpc_client.py
```diff
@@ -467,6 +458,27 @@ def nvmf_subsystem_remove_ns(self, nqn, nsid):
...
+        params = {
+            "nqn": nqn,
+            "nsid": nsid,
+            "bdev_name": bdev_name,
+        }
+        try:
+            return self._request("nvmf_subsystem_ns_update", params)
+        except Exception as e:
+            logger.debug("nvmf_subsystem_ns_update not available or failed: %s", e)
+            return None
```
mxsrc, line 502: Use `_request3` and do not catch the resulting exception here, handling it is the caller's responsibility.

simplyblock_core/rpc_client.py
```diff
@@ -1393,10 +1396,11 @@ def bdev_lvol_set_lvs_read_only(self, lvs_name, read_only=False):
...
 
-    def bdev_lvol_create_hublvol(self, lvs):
-        return self._request('bdev_lvol_create_hublvol', {
-            "uuid" if utils.UUID_PATTERN.match(lvs) else "lvs_name": lvs,
-        })
+    def bdev_lvol_create_hublvol(self, lvs, name=None):
+        params = {"uuid" if utils.UUID_PATTERN.match(lvs) else "lvs_name": lvs}
+        if name:
+            params["name"] = name
+        return self._request('bdev_lvol_create_hublvol', params)
```
mxsrc, line 1446: Use `_request3` here. Also, why change that function at all? There's no change here.

simplyblock_core/storage_node_ops.py
```diff
@@ -7139,7 +7074,7 @@ def auto_repair(node_id, validate_only=False, force_remove_inconsistent=False, f
     # #sbctl sn list-snapshots d4577fa7-545f-4506-b127-7e81fc3a6e34 --json > snaps_8080.json
     # with open('snaps_8082.json', 'r') as file:
     #     snaps = json.load(file)
-    snaps = snapshot_controller.list_by_node(node_id, is_json=True)
+    snaps = snapshot_controller.list_snapshots(node_id=node_id, is_json=True)
     if snaps:
         snaps = json.loads(snaps)
```
mxsrc, line 7334: This looks to be the only reason for introducing the `is_json` parameter. This can simply be pruned and we'd avoid adding that.

simplyblock_web/api/v2/migration.py
```diff
@@ -2,72 +2,105 @@
...
+    try:
+        migration_id, connect_strings = migration_controller.pre_create_on_target(
+            volume.get_id(),
+            str(parameters.target_node_id),
+            ctrl_loss_tmo=parameters.ctrl_loss_tmo,
+            host_nqn=parameters.host_nqn,
+        )
+    except ValueError as e:
+        raise HTTPException(400, str(e))
+    return _PreCreateResponse(migration_id=UUID(migration_id), connect_strings=connect_strings)
```
mxsrc: https://github.com/simplyblock/sbcli/pull/1094 will introduce a consistent way of returning responses for creation of entities. Ideally, this would already follow the same approach. I suggest:

- fold the connect values into the `MigrationDTO` generally
- simply return the `MigrationDTO` here.

simplyblock_web/api/v2/migration.py
```diff
@@ -2,72 +2,105 @@
...
 
 
-@instance_api.post('/cancel', name='clusters:migrations:cancel', status_code=200)
-def cancel_migration(cluster: Cluster, migration: Migration):
-    from simplyblock_core.controllers import migration_controller
-    ok, error = migration_controller.cancel_migration(migration.get_id())
-    if not ok:
-        raise HTTPException(400, error)
+class _ContinueParams(BaseModel):
+    max_retries: int = 10
```
mxsrc: This can't be negative. Use `util.Unsigned`.

simplyblock_web/api/v2/migration.py
```diff
@@ -2,72 +2,105 @@
...
+    deadline_seconds: int = 14400
```
mxsrc: FastAPI supports using `datetime.timedelta`. Might make sense to use here and drop the `_seconds` suffix.

simplyblock_web/api/v2/migration.py
```diff
@@ -2,72 +2,105 @@
...
+@instance_api.post('/continue', name='clusters:storage-pools:volumes:migrations:continue', status_code=200)
+def continue_migration(volume: Volume, migration: Migration, parameters: _ContinueParams):
+    try:
+        migration_id = migration_controller.start_migration(
+            migration_id=migration.get_id(),
+            max_retries=parameters.max_retries,
+            deadline_seconds=parameters.deadline_seconds,
+        )
+    except ValueError as e:
+        raise HTTPException(400, str(e))
```
mxsrc: Above I suggested to consistently use `PreconditionError`. When that's done, there's no need to explicitly handle the exception. The generic handlers will map `PreconditionError`s to a 400 status and anything else to a 500, which is correct.

mxsrc (changes requested):
Thanks for the changes! I have some more comments.

The big one is that the migration task itself in my opinion is not maintainable as it is. The module has multiple functions with very high complexity, indicating structure only through comments. Those should be split up into individual sub functions. Some of them even look similar to other parts of our code, but in the current state it's impossible to tell whether things may be reused, this issue will only become worse.

simplyblock_core/controllers/health_controller.py
EbiRider: i actually don't think if i touched this file but I'll have a look see what's it's trying to accomplish

PR #1100, opened 2026-06-15
Toggle cluster secret auth

mxsrc:
This introduces an option to toggle cluster-secret-based authentication in the API. Since the tokens are long-lived and there is no rotation mechanism for them, depending on security requirements it makes sense for some deployments to completely rely on the service account based authentication introduced recently.

Since the cluster-secret is the only authentication mechanism for API v1, it requires it to be disabled for this purpose. This option is also introduced in this PR, by configuring a list of enabled API versions, which yields a generic option for managing which API versions are available.

### DIFF
--- a/simplyblock_web/api/__init__.py
+++ b/simplyblock_web/api/__init__.py
@@ -1,9 +1,4 @@
-from fastapi import APIRouter
-
 from . import v1
 from . import v2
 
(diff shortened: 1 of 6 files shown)


Hamdy-khader (approved):
Looks good

PR #1101, opened 2026-06-15
fix: add retry logic to NVMe controller attachment in storage_node_ops.py

Hamdy-khader:
Fixes: https://simplyblock.atlassian.net/browse/SFAM-2755


mxsrc:
Just randomly found this, shouldn't this already be handled by the same thing as the RPCClient retries? If the sleep customization is important a simpler solution might be to make the RPCClient retry behavior more configurable and use a one-time client for this task.

PR #1103, opened 2026-06-15
Prevent secret leakage

mxsrc:
The extensive logging done throughout the application may inadvertently leak confidential data. This PR aims to mitigate this, preventing secrets to be present in the logs. 

The core tool in this PR is the use of pydantic's `SecretStr`/`SecretBytes`. They are a simple wrapper around `str`/`bytes`, that, when converted to string yield a redacted string (`**********`). The actual value needs to be explicitly unwrapped via `get_secret_value()`.

This changeset introduces:
- support for using fields of the secret type in models
- unwrapping of secret values in the clients, s.t. the secret values are correctly unwrapped before transmitting them
- Entrypoints in CLI/API wrap confidential values as soon as possible

What htis PR notably does _not_ do, is prevent explicit leakage through endpoints that currently allow access to the confidential values. Handling that in a better way is left as a follow-up.


Hamdy-khader (approved):
Looks good

PR #1104, opened 2026-06-15
feat(node-removal): online node removal as inverse of cluster expansion

schmidt-scaled:
## Summary

Implements an **online storage-node removal** function that works as the inverse of cluster expansion (`add_node`). It is **background, task-driven, idempotent and resumable** — `sn remove` validates preconditions then queues an `FN_NODE_REMOVAL` task; the new `tasks_runner_node_removal` service drives the flow:

```
shutdown → in_removal → rewire LVS replicas → remove/fail/migrate devices → removed
```

This replaces the old `remove_storage_node` semantics (which only handled an already-offline node and did no LVS rewiring).

## Behavior

**Preconditions (enforced before anything is queued):**
- target node is `online`;
- every other non-removed node is `online`;
- FTT headroom allows losing the node (`_check_ftt_allows_node_removal`);
- the node hosts **no LVols and no snapshots** (operator migrates those separately, at a higher level);
- every secondary/tertiary replica the node hosts for *other* primaries has a host-disjoint relocation target (catches e.g. 2-node clusters).

**LVS rewiring:**
- **Case A** — the node's *own* primary LVS: tear down its (empty) secondary/tertiary replicas on the peers that host them, clear cross-references.
- **Case B** — replicas this node hosts *for other primaries*: re-host each on a fresh, anti-affinity-valid node (`get_secondary_nodes` / `get_secondary_nodes_2` + `recreate_lvstore_on_non_leader`) so those primaries keep their fault tolerance. Bookkeeping back-ref on the removed node is cleared only after a successful rebuild, so retries resume cleanly.

**Devices:** each data device is driven `online → removed → failed` (queuing failure-migration on the surviving online nodes), then the flow waits for `failed_and_migrated` before flipping the node to `removed`.

## Changes
- `models/base_model.py`: new `STATUS_IN_REMOVAL` (code 13)
- `models/job_schedule.py`: new `FN_NODE_REMOVAL`
- `controllers/tasks_controller.py`: `add_node_removal_task` + dedup branch + `get_active_node_removal_task`; skip `IN_REMOVAL` nodes when fanning out device-migration tasks (their SPDK is dead)
- `storage_node_ops.py`: rewrite `remove_storage_node` (validate + queue) + `node_removal_orchestrate` + Case A/B + device-decommission/finalize helpers
- `services/tasks_runner_node_removal.py`: new runner service (lease-aware, suspend-and-revisit for the long migration wait)
- `scripts/docker-compose-swarm.yml`: register `TasksNodeRemovalRunner`
- `simplyblock_cli/cli.py`: updated `sn remove` help; `--force-remove` now only cancels active tasks
- `tests/test_node_removal.py`: 21 unit tests

## Testing
- New: `tests/test_node_removal.py` — 21 tests (preconditions, relocation feasibility/picking, Case A/B bookkeeping incl. idempotency/resume, device completion gate, status mapping). ✅
- Full non-migration suite: **1137 passed, 213 skipped**, no regressions.

## Notes / review focus
- **Case B** (recreating a replica on a fresh node while the cluster is online) is the net-new, highest-risk piece — it reuses `recreate_lvstore_on_non_leader` with the primary as the online leader. Worth a careful look from someone close to the LVS restart/recreate machinery.
- Unit tests mock all data-plane RPCs; this has **not** yet been validated on a live cluster.


simplyblock_core/storage_node_ops.py
```diff
@@ -2209,99 +2210,443 @@ def delete_storage_node(node_id, force=False):
...
-            for lvol in lvols:
-                lvol_controller.delete_lvol(lvol.get_id(), True)
-        else:
-            logger.warning("LVols found on the storage node, use --force-remove or --force-migrate")
+        cands = get_secondary_nodes_2(
+            primary, exclude_ids=exclude_ids, exclude_mgmt_ips=exclude_mgmt_ips)
+    return cands[0] if cands else None
+
+
+def node_removal_orchestrate(node_id, force_remove=False):
```
mxsrc, line 3436: I think this should be moved to the task module, including its helpers. The tasks should be as self-contained as possible, this code should not be executed from anywhere else.

simplyblock_core/storage_node_ops.py
```diff
@@ -2209,99 +2210,443 @@ def delete_storage_node(node_id, force=False):
...
-    if lvols:
-        if force_migrate:
-            for lvol in lvols:
+    # Case-B feasibility: every replica this node hosts for another primary must
+    # have somewhere host-disjoint to go. Catches e.g. 2-node clusters where the
+    # tertiary cannot be re-placed without violating anti-affinity.
+    feasible, reason = _check_replica_relocation_feasible(snode, db_controller)
+    if not feasible:
+        logger.error(f"Can not remove node {node_id}: {reason}")
+        return False
```
mxsrc, line 3380: Since we rework this function completely, I think we should convert this from returning a success-boolean to exceptions, in particular for `PreconditionError`.

This also introduces quite a few checks, I'm not sure whether some of them are new. In any case, did we check that after starting this task, none of these conditions can be removed by external factors? I.e. do we consistently prevent
- this node (or any other node) being shut down/suspended/restarted
- LVols/Snapshots being created/moved to this node
- Tasks being scheduled for this node

I think at least for some of them we already have checks, but to avoid race conditions we'd have to at least set the `IN_REMOVAL` status here. Technically, a two phase commit would be needed, setting the status or a lock before running the checks, and clearing it if any fails.

mxsrc (changes requested):
The new functions that are being introduced use return-value based error reporting, since they are new I feel they should use exceptions for the error handling. Other than that I think the code looks good, I just have a few comments on two locations. It requires changes in the helm-chart to add the container there as well.

Hamdy-khader (approved):

tests/unit/test_node_removal.py
```diff
@@ -0,0 +1,405 @@
+# coding=utf-8
```
mxsrc, line 1: This file needs to be sorted into either `tests/unit` or `tests/integration`, otherwise it's silently ignored. It looks to me more like an integration test, in that case instead of mocking the DB, it should rely on the one provided by the integration tests. You can check recent changes to the other integration tests for how this may look.

simplyblock_core/env_var
```diff
@@ -1,5 +1,5 @@
 SIMPLY_BLOCK_COMMAND_NAME=sbcli-dev
 SIMPLY_BLOCK_VERSION=19.2.34
 
-SIMPLY_BLOCK_DOCKER_IMAGE=public.ecr.aws/simply-block/simplyblock:main
+SIMPLY_BLOCK_DOCKER_IMAGE=public.ecr.aws/simply-block/simplyblock:feature-node-removal
```
mxsrc, line 4: This must not be merged to main.

simplyblock_core/services/storage_node_monitor.py
```diff
@@ -434,6 +434,7 @@ def _spawn_recovery_shutdown(node):
 
     def _run():
         try:
+
```
mxsrc, line 660: This is superfluous

simplyblock_core/scripts/docker-compose-swarm.yml
```diff
@@ -484,6 +484,20 @@ services:
     environment:
       SIMPLYBLOCK_LOG_LEVEL: "$LOG_LEVEL"
 
+  TasksNodeRemovalRunner:
```
mxsrc, line 501: We should rename this to `TasksNodeRemoveRunner` for consistency.

simplyblock_core/models/job_schedule.py
```diff
@@ -17,6 +17,7 @@ class JobSchedule(BaseModel):
     FN_FAILED_DEV_MIG = "failed_device_migration"
     FN_NEW_DEV_MIG = "new_device_migration"
     FN_NODE_ADD = "node_add"
+    FN_NODE_REMOVAL = "node_removal"
```
mxsrc, line 20: We should rename this to `FN_NODE_REMOVE = "node_remove"` for consistency.

simplyblock_core/models/nvme_device.py
```diff
@@ -105,6 +105,9 @@ class JMDevice(NVMeDevice):
     # the per-leg member partitions. Empty for single-device (no-raid) JMs.
     jm_leg_bdevs: List[str] = []
     jm_leg_members: List = []
+    # When attaching this JM to a node, override the device name on that node.
+    # This is needed when a JM device is removed and needed to be replaced, but the name must be the same.
+    override_name_on_node: dict[str, str] = {} # node_id: new_name
```
mxsrc, line 133: Ideally we'd provide a single helper `JMDevice.controller_name(node_id: UUID) -> str` that is used to centrally determine the controller name and is called from storage_node_ops. The derivation logic is duplicated and prone to be missed in future changes.

mxsrc (changes requested):
Looks largely good, I think there's some friction with merging to main that should be addressed first.

PR #1105, opened 2026-06-17
Restructure API v2 module

mxsrc:
The flat file structure hid the path hierarchy. This change aligns the two, with some minor differences (lack of ID, singular vs. plural). For this to work, the dependency definitions where extract to a private `_dependencies` module. The DTO module was aligned to be private as well. Semantically, nothing changes with this.


Hamdy-khader (approved):
Looks good

noctarius:
Same. Looks good 😀

PR #1108, opened 2026-06-18
POST /snapshot: return error if value already exists

boddumanohar:
* SnapshotDTO doesn't have `created_at` field although the model returns it. 
* Return 409 during POST /snapshot if the snapshot with the name already exists. 

the upstream function returns the error. so we look for the error:
https://github.com/simplyblock/sbcli/blob/18d1efaf9364d1b91fd45f67d539e213b26dac7b/simplyblock_core/controllers/snapshot_controller.py#L104-L106

### DIFF
--- a/simplyblock_web/api/v2/_dtos.py
+++ b/simplyblock_web/api/v2/_dtos.py
@@ -1,4 +1,4 @@
-from datetime import timedelta
+from datetime import datetime, timedelta
 from ipaddress import IPv4Address
 from typing import List, Literal, Tuple, Optional, cast
 from uuid import UUID
@@ -249,7 +249,8 @@ class SnapshotDTO(BaseModel):
     used_size: util.Unsigned
     migrating: bool
     lvol: Optional[util.UrlPath]
-    created_at: int
+    created_at: datetime
+
 
     @staticmethod
     def from_model(
--- a/simplyblock_web/api/v2/_dtos.py
+++ b/simplyblock_web/api/v2/_dtos.py
@@ -249,6 +249,7 @@ class SnapshotDTO(BaseModel):
     used_size: util.Unsigned
     migrating: bool
     lvol: Optional[util.UrlPath]
+    created_at: int
 
     @staticmethod
     def from_model(
@@ -271,6 +272,7 @@ class SnapshotDTO(BaseModel):
             size=model.size,
             used_size=model.used_size,
             migrating=is_migrating,
+            created_at=model.created_at,
             lvol=str(
                 request.url_for(
                     "clusters:pools:volumes:detail",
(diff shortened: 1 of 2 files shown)


simplyblock_web/api/v2/cluster/storage_pool/volume.py
```diff
@@ -288,6 +288,8 @@ def create_snapshot(
         volume.get_id(), parameters.name, backup=parameters.backup
     )
     if err_or_false:
+        if 'unique' in str(err_or_false).lower():
```
mxsrc, line 293: I really struggle with the thought of matching strings in error messages. I commonly reword those without checking whether they are used anywhere, I feel others do to. I see why you did this though, given the current code. Options I see to addressing this are:
- checking for duplicates before calling the function (expensive, as it is a name-based lookup which our database doesn't handle well)
- rework the add function to use exception-based error handling, return `PreconditionError`s for the precondition checks, introduce a e.g. `NameExistsError` as a `PreconditionError` subclass, and introduce an exception handler that always maps that to a 409 error.

Obviously I prefer the latter.... But of course it's much more involved.

simplyblock_web/api/v2/_dtos.py
```diff
@@ -249,6 +249,7 @@ class SnapshotDTO(BaseModel):
     used_size: util.Unsigned
     migrating: bool
     lvol: Optional[util.UrlPath]
+    created_at: int
```
mxsrc: This should be a `datetime`.

mxsrc (changes requested):
Let me know your thoughts on my first comment. The second one should definitely be addressed though.

simplyblock_web/api/v2/cluster/storage_pool/volume.py
boddumanohar: Wanted to perform a minimally invasive surgery 😅 . but ideally I'd prefer the later too. let check me the uses of this function and reply

simplyblock_web/api/v2/_dtos.py
boddumanohar: fixed in [65f1bdf](https://github.com/simplyblock/sbcli/pull/1108/commits/65f1bdf0092108b319197e46532a823ea0a88673)

simplyblock_web/api/v2/cluster/storage_pool/volume.py
boddumanohar: discussed in slack we'd go with this changes for now. And this can be done in a separate PR later.

simplyblock_web/api/v2/_dtos.py
mxsrc: I think we talked past each other, it should be:
```
created_at: datetime
```
With the conversion converting the timestamp into a proper datetime entity.

boddumanohar: sorry, yeah. fixed it.

mxsrc (approved):
Looks good :)

PR #1111, opened 2026-06-18
fix: add backup encryption key management to KMS and integrate with backup workflows

Hamdy-khader:
Fixes [sfam-2752](https://simplyblock.atlassian.net/browse/SFAM-2752)

### DIFF
--- a/simplyblock_cli/clibase.py
+++ b/simplyblock_cli/clibase.py
@@ -1029,11 +1029,7 @@ class CLIWrapperBase:
             clusters = db.get_clusters()
             if clusters:
                 cluster_id = clusters[0].get_id()
-        success, error = backup_controller.switch_backup_source(
-            cluster_id, args.source_cluster_id)
-        if error:
-            print(f"Error: {error}")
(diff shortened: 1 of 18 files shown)


mxsrc:
I had to write a bit more code to address this. The fundamental reason is the need to re-key the DEKs using a backup-level KEK, s.t. their lifetime is decoupled from the original lvol (DEK) and pool (KEK). In the scope of this change I also
introduced some structure for the DEKs and KEKs that allows to write more restrictive policies if deemed necessary.
DEKs now live in `cluster/$id/lvol/$id` and `backup/$id`, while KEKs are named `cluster/$id/pool/$id` and `backup/$id` respectively. To restore a backup from a foreign cluster, the backup KEKs and DEKs need to be present in the configured KMS. When using the same KMS this is achieved trivially.

mxsrc:
There was a failure where the cluster activation silently ignored errors of the backup component which induced later failures. I adapted the code s.t. this failure will be caught at activation time.

PR #1113, opened 2026-06-19
fix snapshot API response

boddumanohar:
Fixes responses for snapshot API

* return `created_at` in RFC 3339 format
* In the pool-level listing `volume_id` is `None`, so `None == model.lvol.get_id()` is always `False`. meaning `lvol` would be `None` in every pool-level response even when the snapshot has a source volume.

### DIFF
--- a/simplyblock_web/api/v2/_dtos.py
+++ b/simplyblock_web/api/v2/_dtos.py
@@ -1,4 +1,4 @@
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from ipaddress import IPv4Address
 from typing import List, Literal, Tuple, Optional, cast
 from uuid import UUID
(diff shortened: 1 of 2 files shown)


mxsrc:
I'm not sure I totally understand, a bit confused about the reasoning for passing the volume_id in in the first place. Shouldn't we just always use `model.lvol.uuid`?

boddumanohar:
The SnapshotDTO is used at 2 places:

1. clusters:storage-pools:snapshots:list --> volume_id is empty
2. clusters:storage-pools:volumes:snapshots:list --> volume_id is passed. And that line checks if the passed in the volume_id is same as the current context which is in model.lvol.get_uuid()


At the moment, `lvol` is returned empty for pool level snapshots API. To prevent that, I added this extra check

boddumanohar:
It seems like we already have a check for this in: volume.py:278

```
if snapshot.lvol is not None and snapshot.lvol.get_id() == volume.get_id()
```

so I think we can safely remove the volume_Id parameter from from_model function in SnapshotDTO

mxsrc (approved):
Looks good! Thanks for the changes.

PR #1115, opened 2026-06-22
Convert lvol deletion error handling to exceptions

mxsrc:
This converts lvol delete error handling to exceptions. This includes a fix for a CLI bug that caused deletion of multiple volumes to silently ignore any but the first volume. It also changes the signature s.t. the lvol is passed by value instead of ID, avoiding duplicate lookups, since all but one call-sites already retrieved the lvol from the database.

The motivation for this PR was a cleaner error handling through the API when the request was rejected due to an ongoing migration.


Hamdy-khader (approved):
Looks good

PR #1116, opened 2026-06-22
register missing clusters:backups:detail route

boddumanohar:
PR #1094 refactored `create_backup` to use the shared `creation_response()` helper, which generates a `Location` header via `url_path_for(route_name, **kwargs)`. It referenced `route_name='clusters:backups:detail'` but never
registered the matching `GET /{backup_id}` endpoint.


mxsrc (approved):
Good catch, thanks for the fix!

PR #1117, opened 2026-06-22
Fix type check errors

mxsrc:
The type checker complains about the lambda return value being used. Introducing inline functions to bundle the transactions avoids this issue.


boddumanohar (approved):
do you think we should add AGENTS.md or CLAUDE.md file so that the agent keeps in mind that type checks also needs to be fixed.

mxsrc:
Yeah, I think that would be a good idea. A while back I added the tox-verify skill but I feel people do not use that skill if it requires manual effort on their part. I'll raise a PR with agent instructions.

PR #1123, opened 2026-06-22
fix: correct route name from volumes:detail to storage-pools:detail i…

geoffrey1330:
…n pool create response


mxsrc (approved):
Good catch, thanks!

PR #1130, opened 2026-06-23
Deduplicate cloud info

mxsrc:
The functions were duplicated, with only the timeout differing. This makes it simpler to mock the output.

### DIFF
--- a/simplyblock_web/api/internal/storage_node/_cloud_info.py
+++ b/simplyblock_web/api/internal/storage_node/_cloud_info.py
@@ -1,0 +1,65 @@
+import requests
+
+
+def get_cloud_info() -> dict | None:
+    return next((
+        info
+        for getter
+        in [_google_info, _amazon_info, _equinix_info]
+        if (info := getter()) is not None
+    ), None)
+
+
+def _google_info() -> dict | None:
+    try:
+        headers = {'Metadata-Flavor': 'Google'}
+        response = requests.get("http://169.254.169.254/computeMetadata/v1/instance/?recursive=true", headers=headers, timeout=2)
+        data = response.json()
+        return {
+            "id": str(data["id"]),
+            "type": data["machineType"].split("/")[-1],
+            "cloud": "google",
+            "ip": data["networkInterfaces"][0]["ip"],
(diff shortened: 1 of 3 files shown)


simplyblock_web/api/internal/storage_node/_cloud_info.py
```diff
@@ -0,0 +1,65 @@
...
+                if interface["enabled"] and interface["public"]:
+                    public_ip = interface["address"]
+                elif interface["enabled"] and not interface["public"]:
+                    public_ip = interface["address"]
+        return {
+            "id": str(data["id"]),
+            "type": data["class"],
+            "cloud": "equinix",
+            "ip": public_ip,
+            "public_ip": ip
```
boddumanohar: these 2 seems inverted? 

sorry, who are equinix?

boddumanohar: looks like some bare-metal cloud provider

simplyblock_web/api/internal/storage_node/_cloud_info.py
```diff
@@ -0,0 +1,65 @@
+import requests
+
+
+def get_cloud_info() -> dict | None:
+    return next((
+        info
+        for getter
+        in [_google_info, _equinix_info, _amazon_info]
```
boddumanohar: We don't seem to use `_equinix_info` that probably we can move this to the last of the list as it seems to connect to a public IP.

simplyblock_web/api/internal/storage_node/_cloud_info.py
mxsrc: Honestly, no idea. I just copied it as-is.

mxsrc: Fixed the inversion though, even if I can't test it.

simplyblock_web/api/internal/storage_node/_cloud_info.py
mxsrc: Yeah, I guess that makes sense.

boddumanohar (approved):

PR #1132, opened 2026-06-23
fix(restart): retry distrib (re)creation on stale cluster map, else abort

schmidt-scaled:
## Problem
During a node restart, if a peer node goes offline at the exact moment the restarting node rebuilds its lvstore stack, the cluster map is briefly stale (the departed node's devices still flagged online) and `bdev_distrib_create` / the cluster-map push fails. `_create_bdev_stack._create_distr` **swallowed** that failure (logged only) and the loop marked the distrib branch successful regardless — so the restart completed on a broken distrib.

## Fix
- Track each distrib's creation outcome.
- Retry a failed distrib **exactly once** (short delay for the CP to reconcile the departed node, delete any half-created distrib, rebuild a fresh map).
- If the retry also fails, `_create_bdev_stack` rolls back and returns an error — it never assembles the raid on top of a broken distrib — so `recreate_lvstore` aborts the restart and the task runner queues an **auto-restart** (the standard unrecoverable-error path), which succeeds once the map is fresh.

## Tests
`tests/test_distrib_recreate_retry.py`: retry-succeeds-after-stale-map, map-push-failure-then-retry, abort-when-retry-also-fails (asserts no raid built + rollback). Verified against the restart/recreate/dual-FT suites (80 passed / 6 skipped).

### DIFF
--- a/simplyblock_core/storage_node_ops.py
+++ b/simplyblock_core/storage_node_ops.py
@@ -7039,15 +7039,52 @@ def create_lvstore(snode, ndcs, npcs, distr_bs, distr_chunk_bs, page_size_in_blo
 
 
 
+# Seconds to wait before the single retry of a failed distrib (re)creation,
+# giving the control plane time to reconcile a peer that went offline mid-restart
+# so the rebuilt cluster map no longer references its devices as online.
+_DISTR_RECREATE_RETRY_DELAY_SEC = 5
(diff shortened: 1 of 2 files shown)


schmidt-scaled:
Superseded — pushed directly to main (commit 3d8f1a6c).

PR #1135, opened 2026-06-25
Migration followup

EbiRider:
in this follow up we addressing multiple issues for migration 

1) the double m suffix issue from back to back migrations which made subsequent migrations fail 
2) tertiary node not being registered on the target's tertiary node
3) cleanup of the code for unused variables and parameters
4) extracting hublvol control logic out of the migration flow to ensure detach and attach of the hublvol respect it's needed cooldown of ten seconds


simplyblock_core/models/cluster.py
```diff
@@ -58,6 +58,10 @@ class Cluster(BaseModel):
     distr_ndcs: int = 0
     distr_npcs: int = 0
     enable_node_affinity: bool = False
+    # Undocumented chaos-testing switch: when set at cluster create, a transparent
+    # SPDK delay bdev is inserted between each device's nvme bdev and its alceml so
+    # a device can be made to "hang" on demand (see device_controller device-hang).
+    enable_hang_device: bool = False
```
mxsrc, line 73: I think we need to very carefully think about the implications of integrating fault injection into the business logic. Where is this called from? I cannot find it in this PR. Is there no other way of achieving this?

simplyblock_core/controllers/migration_controller.py
```diff
@@ -222,7 +216,8 @@ def _cleanup_created(migration):
         return
 
     nqn = lvol.nqn
-    bdev_short = lvol.lvol_bdev + _MIGRATION_BDEV_SUFFIX
+    _lvol_base = lvol.lvol_bdev[:-len(_MIGRATION_BDEV_SUFFIX)] if lvol.lvol_bdev.endswith(_MIGRATION_BDEV_SUFFIX) else lvol.lvol_bdev
+    bdev_short = _lvol_base + _MIGRATION_BDEV_SUFFIX
```
mxsrc: This is not the only place where this is failing. A proper fix uses the idempotent helper that is already present consistently: https://github.com/simplyblock/sbcli/pull/1167/changes/0a9690a928f5c85463a9e9a1d50d3f7d4be3f7ba.

simplyblock_core/controllers/migration_controller.py
```diff
@@ -666,7 +679,8 @@ def create_migration(lvol_id, target_node_id,
     cluster = db.get_cluster_by_id(tgt_node.cluster_id)
     tgt_rpc = tgt_node.rpc_client()
     nqn = lvol.nqn
-    bdev_short = lvol.lvol_bdev + _MIGRATION_BDEV_SUFFIX
+    _lvol_base = lvol.lvol_bdev[:-len(_MIGRATION_BDEV_SUFFIX)] if lvol.lvol_bdev.endswith(_MIGRATION_BDEV_SUFFIX) else lvol.lvol_bdev
+    bdev_short = _lvol_base + _MIGRATION_BDEV_SUFFIX
```
mxsrc: This is too easy to miss, same comment as above. If you repeat code, use a helper.

mxsrc (commented):
I'm a bit lost with the PR, the description is not really enough for me to understand what is going on. Ideally we'd split the PR up into indivudal changes, not sure whether that's feasible. Currently the description says the PR does four things, but there is definitely more going on. Questions I have from a quick glance that I can't really answer, that don't seem to correspond to any of the items the description indicates:
- What are the test scripts that are being introduced for, who is supposed to execute them?
- What's the purpose of the fault injection logic, where is it used from?
- Why does the PR introduce replication logic, my understanding is that it's supposed to fix volume migration

Please help me review this, as it stands I don't understand what is going on.

simplyblock_core/rpc_client.py
```diff
@@ -1309,6 +1309,10 @@ def bdev_lvol_get_lvstores(self, name):
         params = {"lvs_name": name}
         return self._request("bdev_lvol_get_lvstores", params)
 
+    def bdev_lvol_rename(self, old_name, new_name):
+        params = {"old_name": old_name, "new_name": new_name}
+        return self._request("bdev_lvol_rename", params)
```
mxsrc: New RPCs should use `_request3`.

simplyblock_core/services/tasks_runner_lvol_migration.py
```diff
@@ -688,7 +614,8 @@ def _cleanup_final_migration(src_rpc, ctx, tgt_rpc=None, rollback_target=False):
                 logger.warning(f"cleanup target subsystem {nqn}: {e}")
         if tgt_composite and tgt_rpc.get_bdevs(tgt_composite):
             try:
-                _delete_bdev_blocking(tgt_composite, tgt_rpc)
+                _delete_bdev_blocking(tgt_composite, tgt_rpc,
+                                      secondary_rpc=tgt_sec_rpc, tertiary_rpc=tgt_ter_rpc)
             except Exception as e:
                 logger.warning(f"cleanup target lvol {tgt_composite}: {e}")
```
mxsrc, line 663: Shouldn't this fail if the deletion fails? Currently it doesn't.

simplyblock_core/controllers/migration_controller.py
```diff
@@ -547,7 +560,7 @@ def _collect_snap_ancestry(snap_uuid, out_set):
         out_set.add(current)
         try:
             snap = db.get_snapshot_by_id(current)
-            current = snap.snap_ref_id
+            current = snap.snap_ref_id or snap.prev_snap_uuid
         except KeyError:
             break
```
mxsrc: This, and `_get_snap_ancestry` seem redundant. Is there no way of deduplicating them?
At least `_protect_snap_and_ancestors` can basically be `candidate_set - _collect_snap_ancestry()`, can't it?

simplyblock_core/services/tasks_runner_lvol_migration.py
```diff
@@ -398,18 +341,28 @@ def _snap_short_name(snap):
 def _snap_tgt_short_name(snap):
```
mxsrc, line 336: This is the helper that should be used anywhere the value needs to be used.

simplyblock_core/services/tasks_runner_lvol_migration.py
```diff
@@ -461,39 +414,44 @@ def _log_spdk_bdev_size(rpc, composite_name, label):
...
+    deadline = time.monotonic() + timeout_s
+    while True:
         status = primary_rpc.bdev_lvol_get_lvol_delete_status(bdev_name)
-        if status in (0, 2):
-            break
-        if status == 1:
-            time.sleep(0.25)
-        else:
-            logger.warning(f"delete bdev {bdev_name}: unexpected status {status}")
+        if status != 1:
```
mxsrc: This can be folded into the loop condition.

mxsrc (changes requested):
There's a few minor things that should change, I left comments for that. The two larger things are that quite a few places ignore errors instead of escalating them, sometimes logging a warning, sometimes not. In some places this doesn't seem right, I think this should be revisited. As for the core logic, I can't really judge that, this, and the pre-existing code doesn't really document the intent, but from a code-perspective I think we can go forward.

simplyblock_core/rpc_client.py
EbiRider: done, using _request3 for the rpc now

simplyblock_core/services/tasks_runner_lvol_migration.py
EbiRider: throwing an exception runtime error now

simplyblock_core/controllers/migration_controller.py
EbiRider: fixed

simplyblock_core/services/tasks_runner_lvol_migration.py
EbiRider: i looked through the code and updated every call site i seen

simplyblock_core/services/tasks_runner_lvol_migration.py
EbiRider: done

simplyblock_core/services/tasks_runner_lvol_migration.py
```diff
@@ -924,18 +868,70 @@ def _handle_snap_copy(migration, src_node, tgt_node, src_rpc, tgt_rpc):
...
+            if (_snap_uuid in migration.snaps_migrated
+                    or _snap_uuid in migration.snaps_preexisting_on_target):
+                continue
+            try:
+                _s = db.get_snapshot_by_id(_snap_uuid)
+            except KeyError:
+                continue
+            _short_tgt = _snap_tgt_short_name(_s)          # SNAP_Xm  (in-flight)
+            _short_canonical = _snap_short_name(_s)         # SNAP_X   (post-rename)
+            _short_am = _short_canonical + _MIGRATION_BDEV_SUFFIX_DONE  # SNAP_Xam (fallback)
```
mxsrc, line 976: This will hit the same fault scenario we had with the `BDEV_SUFFIX`, won't it? If the `m` suffix is already applied, simply adding this will make it `SNAP_Xmam`?

simplyblock_core/services/tasks_runner_lvol_migration.py
```diff
@@ -1712,9 +1799,214 @@ def _get_secondary_rpc(node):
...
+                    r = rpc.bdev_lvol_rename(old_composite, new_short)
+                    logger.debug(f"_do_rename {role}: {old_composite!r} -> {new_short!r}: ret={r!r}")
+                    if (not r) or r == _EXISTS:
+                        exists = True
+                except Exception as exc:
+                    logger.warning(
+                        f"_rename_migrated_bdevs: {role} rename {label} (non-fatal): {exc}")
+        if exists:
+            return _EXISTS
+        return True
```
mxsrc, line 1889: I'm struggling to follow this function. It returns either a boolean (or rather, always `True`), or some string sentinel. It also adds relevant logic to `RPCClient.bdev_lvol_rename`. The comments are a bit unclear to me and violate abstraction layers, the RPCClient should contain any knowledge about RPC workings, like the mapping of error codes.

So the suggestion is to add an `exists_ok` flag to `RPCClient.bdev_lvol_rename`, and raise an RPCException subclass that indicates the entity already exists. With that primitive, we can lose the sentinel, and `_rename_with_fallback` becomes much more readable.

In addition to handling the RPC logic, it does the rename on all three nodes that are responsible for the LVol. It doesn't handle inconsistencies or failures correctly though open questions are:
- what happens if the function encounters an existing inconsistency where the target exists on only one of the nodes. There is no way to handle this easily, but at least we need to be aware of it
- Errors of any of the rename (except the first) are ignored and flagged as non-fatal. Is that correct? It introduces an inconsistency that follow up code has to deal with.

simplyblock_core/services/tasks_runner_lvol_migration.py
```diff
@@ -1712,9 +1799,214 @@ def _get_secondary_rpc(node):
...
+    bdev_lvol_rename returns JSON-RPC error -32602 "File exists" (None to caller)
+    when the target name is already taken; we use that to try the fallback
+    (_MIGRATION_BDEV_SUFFIX_DONE) instead of doing a pre-query of all bdevs.
+
+    Must be called AFTER _apply_migration_to_db() — snap.snap_bdev and
+    lvol.lvol_bdev already point to the 'm'-suffixed target paths at that point.
+    """
+    if _SKIP_RENAME_BDEVS:
+        logger.info("_rename_migrated_bdevs: skipped (SKIP_RENAME_BDEVS=True)")
+        return
```
mxsrc: This seems like a debugging provision? In that case I prefer not to mainline it.

mxsrc (commented):
Gotta look into something else and didn't finish the review, leaving my comments here for now. Something I noticed, not necessarily for now, but a common pattern is to do an operation on a node, as well as their secondary and primary. One single helper would help doing this correctly, ensuring correct error handling.

simplyblock_core/services/tasks_runner_lvol_migration.py
mxsrc: https://github.com/simplyblock/sbcli/pull/1135/changes/88bda5b19d071f516af0e76386540b0e320d3405#r3537839105

simplyblock_core/controllers/migration_controller.py
```diff
@@ -529,40 +529,31 @@ def get_snaps_to_delete_on_target(migration):
...
-        visited.add(current)
-        out_set.add(current)
-        try:
-            snap = db.get_snapshot_by_id(current)
-            current = snap.snap_ref_id or snap.prev_snap_uuid
-        except KeyError:
-            break
+def _protect_snap_and_ancestors(snap_uuid, candidate_set):
+    """Remove *snap_uuid* and all its ancestors from *candidate_set*."""
+    candidate_set -= _collect_snap_ancestry(snap_uuid)
```
mxsrc: At this point we don't need the helper anymore, the operation can simply be used at the call-site.

simplyblock_core/services/tasks_runner_lvol_migration.py
mxsrc: https://github.com/simplyblock/sbcli/pull/1135/changes/88bda5b19d071f516af0e76386540b0e320d3405#r3537842139

simplyblock_core/services/tasks_runner_lvol_migration.py
```diff
@@ -1712,9 +1799,214 @@ def _get_secondary_rpc(node):
...
+            except Exception as exc:
+                logger.warning(
+                    f"_rename_migrated_bdevs: ancestor {inst_short} failed: {exc}")
+                continue
+            if target:
+                inst['snap_bdev'] = f"{lvstore}/{target}"
+                updated = True
+
+        if updated:
+            snap.write_to_db(db.kv_store)
```
mxsrc: This is an exact duplicate of the code above.

simplyblock_core/services/tasks_runner_lvol_migration.py
mxsrc: Great, but this needs to be handled at the call-sites as well. Many of the call-site use non-exception-based error handling, so the error needs to be handled appropriately.

simplyblock_core/services/tasks_runner_lvol_migration.py
```diff
@@ -949,7 +945,22 @@ def _handle_snap_copy(migration, src_node, tgt_node, src_rpc, tgt_rpc):
                         return False, True, _WAIT
                     if tgt_sec:
                         sec_rpc = _make_rpc(tgt_sec)
-                    break  # one check is enough
+                elif snap.lvol.ha_type == "ha3":
+                    tgt_sec, sec_err = _get_target_secondary_node(tgt_node)
```
mxsrc: This supports an optional parameter indicating the source ID to support the use-case of migrating away from a node that is draining. This is not passed here. I suspect this was missed because this path is undocumented, it should be explicitly mentioned in the helpers.

simplyblock_core/services/tasks_runner_lvol_migration.py
```diff
@@ -949,7 +945,22 @@ def _handle_snap_copy(migration, src_node, tgt_node, src_rpc, tgt_rpc):
...
+                    if sec_err:
+                        migration.error_message = sec_err
+                        migration.write_to_db(db.kv_store)
+                        return False, True, _WAIT
+                    if tgt_sec:
+                        sec_rpc = _make_rpc(tgt_sec)
+                    tgt_ter, ter_err = _get_target_tertiary_node(tgt_node)
```
mxsrc: This supports an optional parameter indicating the source ID to support the use-case of migrating away from a node that is draining. This is not passed here. I suspect this was missed because this path is undocumented, it should be explicitly mentioned in the helpers.

simplyblock_core/services/tasks_runner_lvol_migration.py
```diff
@@ -1162,33 +1165,49 @@ def _handle_snap_copy(migration, src_node, tgt_node, src_rpc, tgt_rpc):
 
         tgt_sec = None
         sec_rpc = None
-        if snap.lvol.ha_type == "ha":
-            tgt_sec, sec_err = _get_target_secondary_node(tgt_node, src_node.get_id())
+        tgt_ter = None
+        ter_rpc = None
+        if snap.lvol.ha_type in ("ha", "ha3"):
+            tgt_sec, sec_err = _get_target_secondary_node(tgt_node)
```
mxsrc: This supports an optional parameter indicating the source ID to support the use-case of migrating away from a node that is draining. This is not passed here. I suspect this was missed because this path is undocumented, it should be explicitly mentioned in the helpers.

simplyblock_core/services/tasks_runner_lvol_migration.py
```diff
@@ -1162,33 +1165,49 @@ def _handle_snap_copy(migration, src_node, tgt_node, src_rpc, tgt_rpc):
...
             if sec_err:
                 migration.error_message = sec_err
                 migration.write_to_db(db.kv_store)
                 return False, True, _WAIT
             if tgt_sec:
                 sec_rpc = _make_rpc(tgt_sec)
+        if snap.lvol.ha_type == "ha3":
+            tgt_ter, ter_err = _get_target_tertiary_node(tgt_node)
```
mxsrc: This supports an optional parameter indicating the source ID to support the use-case of migrating away from a node that is draining. This is not passed here. I suspect this was missed because this path is undocumented, it should be explicitly mentioned in the helpers.

simplyblock_core/services/tasks_runner_lvol_migration.py
mxsrc: The `STATUS_CUTOVER` was removed, I assume this is outdated?

simplyblock_core/services/hub_controller_manager.py
```diff
@@ -0,0 +1,329 @@
...
+    the task runner retries on the next loop iteration (~3 s) until the gap passes.
+
+Public API (called by tasks_runner_lvol_migration)
+--------------------------------------------------
+  hub_manager.acquire(src_node_id, src_rpc, tgt_node, trtype)
+      → attach (or reuse) the controller, refresh the idle timer
+      → returns (ctrl_name, hub_bdev, error)
+
+  hub_manager.release(src_node_id, tgt_node_id)
+      → no-op kept for call-site compatibility; idle timer is driven by acquire()
```
mxsrc: The function does not exist, not even as a no-opt.

simplyblock_core/services/tasks_runner_lvol_migration.py
```diff
@@ -752,11 +678,20 @@ def _setup_snap_transfer(snap, snap_index, migration, src_node, tgt_node,
             _delete_bdev_blocking(tgt_composite, tgt_rpc, sec_rpc)
             return None, f"bdev_lvol_register on secondary failed for snap {snap_uuid}"
         sec_registered = True
-
-    # Helper: clean primary and secondary (if registered) on error
+        if tgt_ter and ter_rpc:
+            ret_ter = ter_rpc.bdev_lvol_register(
+                snap_short, tgt_node.lvstore, snap_uuid_on_tgt, snap_blobid,
+                snap.lvol.lvol_priority_class if hasattr(snap, 'lvol') else 0)
```
mxsrc: The `hasattr` check is inconsistent. In the start we check for attribute presence and truethyness, later we only check the attribute. Why is this actually neccessary, can we proceed if the lvol is not set?

I'd prefer to resolve this once (`lvol = ...`), document why we can continue if it is unknown, and go ahead.

simplyblock_core/services/storage_node_monitor.py
```diff
@@ -216,7 +216,9 @@ def get_next_cluster_status(cluster_id):
             else:
                 node_offline_devices += 1
 
-        if node_offline_devices > 0 or (node_online_devices == 0 and node.status != StorageNode.STATUS_REMOVED):
+        if (node_offline_devices > 0
+                or (node_online_devices == 0 and node.status != StorageNode.STATUS_REMOVED)
+                or node.status == StorageNode.STATUS_OFFLINE):
```
mxsrc, line 326: This doesn't seem to be related to the other changes, why is it here?

simplyblock_core/services/tasks_runner_lvol_migration.py
```diff
@@ -1681,16 +1751,33 @@ def _delete_intermediate_snaps_on_target(migration, tgt_rpc, tgt_sec_rpc=None):
...
+                        try:
+                            rpc.delete_lvol(tgt_composite, del_async=True, special_delete=False)
+                        except Exception as e:
+                            logger.warning(
+                                f"Could not finalize intermediate snap delete {tgt_composite} "
+                                f"on {label}: {e}")
+                    logger.info(f"Deleted intermediate snap bdev {tgt_composite} from target")
             except Exception as e:
                 logger.warning(
                     f"Could not delete intermediate snap {tgt_composite} from target: {e}")
```
mxsrc, line 1842: This can delegate to the `_delete_ bdev_blocking` helper if we add a `coalescing` flag to that. There is also an inconsistency where this logic does not poll, which seems unintentional to me.

simplyblock_core/services/tasks_runner_lvol_migration.py
```diff
@@ -115,6 +116,10 @@
 _INTERMEDIATE_POLL_INTERVAL_S = 1      # seconds between stat checks
 _INTERMEDIATE_POLL_MAX = 300           # max iterations ≈ 5 min
 
+_SKIP_CLEANUP_SOURCE = False
+_SKIP_INTERMEDIATE_SNAP_DELETE = False
+_SKIP_RENAME_BDEVS = False
+
```
mxsrc: These should be removed, along with the logic that uses them.

simplyblock_core/services/tasks_runner_lvol_migration.py
```diff
@@ -1576,14 +1636,20 @@ def _flip(rpc, ip, port, trtype, state, label):
...
                     _swap_namespace(tgt['rpc'], nqn, tgt_ns_bdev,
                                     lvol.uuid, lvol.guid, tgt['node_id'][:8])
-                except Exception as e:
-                    logger.error(f"Namespace swap on {tgt['node_id'][:8]} failed: {e}")
-
-        # Step 5: all TGT paths → correct ANA state at TGT port
-        for i, tgt in enumerate(tgt_paths):
-            state = "optimized" if i == 0 else "non_optimized"
+                except Exception:
+                    raise
```
mxsrc: If we re-raise any exception, the `try-except` can be removed.

mxsrc (changes requested):
Thanks for the changes, quite a few good additions in here. I think we need one more iteration though.

The situation with the `_get_target_{secondary,tertiary}_node` should be handled by:
- adding documentation for why in some cases a suspended node should be returned
- passing the source node ID parameter consistently
- removing the default to enforce this for future changes
- reviewing the `_get_{secondary,tertiary}_rpc` helpers: This needs to mirror the logic for the node get helpers, keeping this in sync is evidently hard. Maybe they should be removed, and instead use `_make_rpc(_get_{secondary_target_tertiary_node(...))`. If they are used for source nodes as well, probably we best introduce separate helpers for that indicating them.

simplyblock_core/services/tasks_runner_lvol_migration.py
EbiRider: yeah, i removed it as it no served a purpose

simplyblock_core/services/tasks_runner_lvol_migration.py
EbiRider: removed it

simplyblock_core/services/tasks_runner_lvol_migration.py
EbiRider: used proper handling like the other sites for the sites that didn't have it, i don't think i spotted more occurrences

simplyblock_core/services/hub_controller_manager.py
EbiRider: removed it

simplyblock_core/services/tasks_runner_lvol_migration.py
EbiRider: done, incorporated it into the helper with coalescing flag

simplyblock_core/services/tasks_runner_lvol_migration.py
EbiRider: removed this as well

simplyblock_core/services/tasks_runner_lvol_migration.py
mxsrc: Did you also make the Parameter non-optional and reviewed the other helpers for whether they violate the consteaint as suggested?

EbiRider: yes src_node_id is now a required parameter instead of a default on both helpers, so a call-site that omits it fails  instead of silently skipping overlap-drain handling

mxsrc (approved):
Looks good, let's go forward. Some of the suggestions are still valid, but they shouldn't block merging this.

simplyblock_core/services/tasks_runner_lvol_migration.py
EbiRider: after migration we have a rename operation, which should bring the lvols back to the same name they had on source on the target unless there's a different lvol with exact same name so I'm am suffix for those cases 
that only happnes if there's an unrelated lvol with the same lvol on the target

simplyblock_core/services/tasks_runner_lvol_migration.py
EbiRider: this i been working on mainly on migration-followup2 branch which it changed I'll save this comment for that branch

PR #1142, opened 2026-06-26
Main reduce hp

wmousa:


geoffrey1330 (approved):

simplyblock_web/api/internal/storage_node/kubernetes.py
```diff
@@ -627,6 +672,42 @@ def get_firewall():
     return utils.get_response(False, "deprecated bath get snode/get_firewall")
 
 
+class UpdateNodeConfigParams(BaseModel):
+    max_lvol: int
+    huge_page_memory: int
+    numa: int
+    ssd_list: list
```
mxsrc: This can be typed better I feel. We should at least use `List[str]` for the ssd_list. In other places of the API we use `ssd_pcie`, does this list also contain a list of PCIe addresses? Then I'd argue we should name this consistenly. For the other values we should at least use `util.Unsigned` as the values cannot be negative. It might also make sense to make the values optional to allow partial updates.

simplyblock_core/env_var
```diff
@@ -1,5 +1,5 @@
 SIMPLY_BLOCK_COMMAND_NAME=sbcli-dev
 SIMPLY_BLOCK_VERSION=19.2.34
 
-SIMPLY_BLOCK_DOCKER_IMAGE=public.ecr.aws/simply-block/simplyblock:main
-SIMPLY_BLOCK_SPDK_ULTRA_IMAGE=public.ecr.aws/simply-block/ultra:main-latest
+SIMPLY_BLOCK_DOCKER_IMAGE=simplyblock/simplyblock:main
+SIMPLY_BLOCK_SPDK_ULTRA_IMAGE=simplyblock/spdk:main-latest
```
mxsrc: This should not be merged.

simplyblock_web/api/internal/storage_node/kubernetes.py
```diff
@@ -627,6 +672,42 @@ def get_firewall():
...
+        node_config["huge_page_memory"] = body.huge_page_memory
+        matched = True
+        break
+
+    if not matched:
+        return utils.get_response(False, "No matching node found for given numa and ssd_list")
+
+    core_utils.store_config_file(node_info, constants.NODES_CONFIG_FILE)
+    return utils.get_response(True)
+
```
mxsrc: This duplicates the docker endpoint, we should define it once and reuse it. Check the bottom of the file for how that's done.

mxsrc (changes requested):
Largely this looks good. A few comments:

I'd like to use a consistent naming of `numa_node`. Instead of `node` or `numa`. "node" in this codebase commonly refers to storage nodes, I feel we should avoid any kind of ambiguity.

As for the general design, I think the hugepages config/state file(s) could be simpler represented in a json file, that would make things a bit more self-descriptive, allow to validate the values, and make the logic more readable. Not sure you have the capacity for that sort of change right now though.

Failures in storing the hugepage allocation are ignored, I feel they should be exposed to the caller for them to decide how to handle them.

What's the difference between the new `update_node_config` endpoint and the old `apply_config`? It seems like there is overlap, at least concerning the intent of both. Can they be merged?

mxsrc:
> What's the difference between the new `update_node_config` endpoint and the old `apply_config`? It seems like there is overlap, at least concerning the intent of both. Can they be merged?

Any comment on this? As-is, we are basically replicating two configuration mechanisms. Or at least it looks like that. Ideally we'd only have one, but if there's a justified reason for having two it should be made clear by the name, and optionally documentation what the purpose of each is.

wmousa:
> > What's the difference between the new `update_node_config` endpoint and the old `apply_config`? It seems like there is overlap, at least concerning the intent of both. Can they be merged?
> 
> Any comment on this? As-is, we are basically replicating two configuration mechanisms. Or at least it looks like that. Ideally we'd only have one, but if there's a justified reason for having two it should be made clear by the name, and optionally documentation what the purpose of each is.

Oh sorry @mxsrc , I forgot to reply, the apply config is only for k8s to set the hugepages on the system before starting the spdk pod, for the update_node_config, it's actually updating the sn_config_file in case we change the max lvol and the hugepages for the node during the restart

noctarius:
> Oh sorry @mxsrc , I forgot to reply, the apply config is only for k8s to set the hugepages on the system before starting the spdk pod, for the update_node_config, it's actually updating the sn_config_file in case we change the max lvol and the hugepages for the node during the restart

Maybe "persist_node_config" would be more obvious :)

mxsrc:
> > > What's the difference between the new `update_node_config` endpoint and the old `apply_config`? It seems like there is overlap, at least concerning the intent of both. Can they be merged?
> > 
> > 
> > Any comment on this? As-is, we are basically replicating two configuration mechanisms. Or at least it looks like that. Ideally we'd only have one, but if there's a justified reason for having two it should be made clear by the name, and optionally documentation what the purpose of each is.
> 
> Oh sorry @mxsrc , I forgot to reply, the apply config is only for k8s to set the hugepages on the system before starting the spdk pod, for the update_node_config, it's actually updating the sn_config_file in case we change the max lvol and the hugepages for the node during the restart

Thanks for the clarification, but this is not obvious to me from reading `update_node_config` and `apply_config`. A reader might interpret that the first writes a config file and the second loads that config or something. It think we need more appropriate naming here, maybe documentation, or find a way to integrate both opreations.

mxsrc (approved):

mxsrc (approved):
Looks good, just rebase on top of main please and see whether that fixes the failing CI :)

PR #1152, opened 2026-06-29
Fix SFAM 2776

mxsrc:
The `LocalKMS` dummy expected the wrong return type from fdb, this addresses the problem. Fix has been confirmed by @geoffrey1330 , and includes tests to catch regression.


mxsrc:
Just noticed the test failures, they seem to be unrelated but exposed by an expansion of the integration tests in this PR. I'll investigate.

noctarius (approved):

PR #1158, opened 2026-07-01
Fix SFAM-2774

Hamdy-khader:
- Remove any dependency to no-param call to RPCS (bdev_get_bdevs,nvmf_get_subsystems,bdev_get_iostat)
- Increase lvol stat collector interval to 30s
- Adds device.pt_bdev_uuid to identify remote bdevs
- Do not check for lvol bdev in the lvol health check service, just check subsystem, listener, namespace


simplyblock_core/controllers/health_controller.py
```diff
@@ -286,7 +271,7 @@ def _check_node_hublvol(node: StorageNode, node_bdev_names=None, node_lvols_nqns
     return passed
 
 
-def _check_sec_node_hublvol(node: StorageNode, node_bdev=None, node_lvols_nqns=None, auto_fix=False, primary_node_id=None) -> bool:
+def _check_sec_node_hublvol(node: StorageNode, node_bdevssss=None, node_lvols_nqnsddd=None, auto_fix=False, primary_node_id=None) -> bool:
```
mxsrc: This change seems to be a typo, or was the intent to remove the parameters? They probably can be dropped now, if I understand the changes correctly.

simplyblock_core/controllers/health_controller.py
```diff
@@ -481,7 +438,7 @@ def _check_sec_node_hublvol(node: StorageNode, node_bdev=None, node_lvols_nqns=N
 
 
 def _check_node_lvstore(
-        lvstore_stack, node, auto_fix=False, node_bdev_names=None, stack_src_node=None) -> bool:
+        lvstore_stack, node, auto_fix=False, node_bdev_namesss=None, stack_src_node=None) -> bool:
```
mxsrc, line 441: This seems to be a typo. Do we want to use the parameter alltogether? It doesn't appear to be used anymore.

simplyblock_core/storage_node_ops.py
```diff
@@ -1458,7 +1445,7 @@ def reconnect_dropped_remote_devs(this_node: StorageNode, node_bdev_names=None):
             try:
                 remote_bdev = connect_device(
                     f"remote_{dev.alceml_bdev}", dev, fresh_node,
-                    bdev_names=node_bdev_names, reattach=False)
+                    bdev_names=[], reattach=False)
```
mxsrc: Maybe an opportunity to drop the parameter from the function?

mxsrc (changes requested):
I like this a lot, there are a few typos in the changeset at the moment.

mxsrc (approved):
Looks good, thanks!

PR #1172, opened 2026-07-10
Enforce retry limit on backup tasks

mxsrc:
This PR fixes https://simplyblock.atlassian.net/browse/SFAM-2785. The backup retry logic did not enforce the maximum retry limit, this is addressed by the initial commit. The follow up commits introduce testing to avoid this issue for other task runners where possible, and converts the `simplyblock_core.services` submodules to not execute at import time, allowing them to be tested individually.

Test failures seem to be coming from upstream, there's no real way to confirm this though.


simplyblock_core/services/new_device_discovery.py
Hamdy-khader, line 1: This file is not used and should be removed

Hamdy-khader (commented):
Looks good

PR #1210, opened 2026-08-03
Fix/cluster activate stranded secondary node

wmousa:


wmousa:
merged on another PR https://github.com/simplyblock/sbcli/pull/1280

PR #1219, opened 2026-08-06
cleanup fix: failing sync delete on nodes after async delete, rename fix 

EbiRider:
a fail at executing the sync delete on one of the nodes has shown to lead to an issue later on
adding retry attempts with sleep in between in case of a failure and logging to trace the issue if it occurs again 
also removing redundant setting of migration flag before cleanup as special delete already handle that

fixing an issue with rename where it would not take the path for the failover name ( should attempt to rename migration artifacts from [name_m] to [name] and if that fails due to existing name should be renamed to [name_am] instead)

### DIFF
--- a/simplyblock_core/services/tasks_runner_lvol_migration.py
+++ b/simplyblock_core/services/tasks_runner_lvol_migration.py
@@ -2257,7 +2257,8 @@ def _delete_intermediate_snaps_on_target(migration, tgt_rpc, tgt_sec_rpc=None, t
             logger.warning(f"Could not remove intermediate snap {snap_uuid} from DB: {e}")
 
 
-def _rename_migrated_bdevs(migration, tgt_node, tgt_rpc, tgt_sec_rpc=None, tgt_ter_rpc=None):
+def _rename_migrated_bdevs(migration, tgt_node, tgt_rpc, tgt_sec_rpc=None, tgt_ter_rpc=None,
+                            warnings=None):
     """
     After migration completes, rename 'm'-suffixed bdevs on the target back to
     their canonical names (without the suffix).  This prevents suffix accumulation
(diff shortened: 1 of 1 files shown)


simplyblock_core/services/tasks_runner_lvol_migration.py
```diff
@@ -2257,28 +2273,37 @@ def _rename_migrated_bdevs(migration, tgt_node, tgt_rpc, tgt_sec_rpc=None, tgt_t
...
-        carry the bdev at the canonical name, so a None return there must not
+        carry the bdev at the canonical name, so a collision there must not
         mask a successful primary rename.
         """
-        ret = tgt_rpc.bdev_lvol_rename(old_composite, new_short)
+        try:
+            ret = tgt_rpc.bdev_lvol_rename(old_composite, new_short)
+            prim_exists = False
+        except RPCException as exc:
+            if exc.code == -32602:
```
mxsrc: I introduced `RPCErrorCode` recently, compare against that please.

simplyblock_core/services/tasks_runner_lvol_migration.py
```diff
@@ -493,10 +493,25 @@ def _async_delete(leader):
...
+                    f"delete bdev {bdev_name} sync finalize failed on replica "
+                    f"(attempt {attempt}/3): {e}"
+                )
+                if attempt < 3:
+                    time.sleep(1)
+        if last_exc is not None:
+            logger.error(
+                f"delete bdev {bdev_name} sync finalize STILL failing after 3 attempts "
+                f"(non-fatal, blob metadata may not be cleared on this replica): {last_exc}"
+            )
```
mxsrc, line 520: This used to be a warning, is that intentional? The print claims this is non-fatal...

mxsrc (changes requested):
Two minor changes I'd like to see here, otherwise this looks good.

mxsrc:
Also the slow integration tests are failing with this, they should be green, please check whether this is introduced here and fix if it is.

EbiRider:
I'll have to check if the integration tests are failing because of my changes

simplyblock_core/services/tasks_runner_lvol_migration.py
EbiRider: well i put these logs because this is typically not a call i would expect to fail on a healthy node but during the tests this call got timedout, never reaching the spdk side so I just wanna confirm that retrying solve the issue if a timeout occurs or not, we don't know the actual reason behind it as of yet

mxsrc (approved):

PR #1220, opened 2026-08-06
Fix integration tests

mxsrc:
This fixes the defunct integration tests:
- adapts tests to diverging upstream development
- introduces default timeouts to avoid excessive execution times on individual hangs. If tests do in fact exceed the timeouts, they can be overridden per test if necessary after identifying that they are in fact supposed to take that much time. Execution times of the 10 longest tests are printed when executing the tests.
- Introduces a cluster state `IN_SHRINK` to fix an LVS unblock on metadata activity during a shrink operation.
- Updates the `RPCClient` to wrap all exceptions that can be expected in `RPCException`


Hamdy-khader (approved):
Looks good

PR #1223, opened 2026-08-06
fix(port_block): add support for FirewallClient based on SPDK version

Hamdy-khader:
Adds a `spdk_version` field to `StorageNode`, derived from the tag of `spdk_proxy_image` and set on node add/restart.

`port_block.set_port()` now checks this version up front: nodes running `R26.2-PRE-latest`, or nodes with no version recorded yet (e.g. nodes that came from an upgrade), go straight to `FirewallClient.firewall_set_port (iptables)` rather than attempting the SPDK `nvmf_port_block/nvmf_port_unblock` RPC first and falling back on a `method-not-found` error.

### DIFF
--- a/simplyblock_core/models/storage_node.py
+++ b/simplyblock_core/models/storage_node.py
@@ -163,6 +163,8 @@ class StorageNode(BaseNodeObject):
     lvol_poller_mask: str = ""
     spdk_proxy_image: str = ""
     transfer_hublvol: HubLVol = None  # type: ignore[assignment]
+    # spdk image tag
+    spdk_version: str = "main"
 
     def get_lvol_subsys_port(self, lvs_name=None):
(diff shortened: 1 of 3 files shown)


mxsrc (commented):
I'm wondering what the problem with the existing approach is. It will print errors because of the unknown methods but is robust to e.g. versions even lower than the given one, image tags that are pinned via `@` and differently named images. Also this leaves the other method, shouldn't we replace it here?

Hamdy-khader:
> I'm wondering what the problem with the existing approach is. It will print errors because of the unknown methods but is robust to e.g. versions even lower than the given one, image tags that are pinned via `@` and differently named images. Also this leaves the other method, shouldn't we replace it here?

the issue is that the method will be found on both version but it will block TCP and RDMA traffix on main but on R26.2.8 it will block only RDMA traffic

PR #1224, opened 2026-08-07
Add lblk cluster mode (Linux block devices via SPDK AIO bdevs) + campaign fixes

schmidt-scaled:
## lblk cluster mode: Linux block devices as storage via SPDK AIO bdevs

Rebased onto current main (152613fa); the restart-claim fix formerly carried on this branch is already on main (17ebddba) and dropped out of the diff. Four commits:

### 1. lblk cluster mode (7757cd20)

Cluster-global `device_mode: nvme|lblk` chosen at `cluster create`. In `lblk` mode, eligible Linux block devices (whole, unmounted, unheld, unpartitioned disks — partitioned only with `--force`/`--force-format`) are wrapped in SPDK AIO bdevs; everything from alceml upward is unchanged, and the inter-node fabric (nvme-tcp/rdma) is untouched.

- **Selection** at `sn configure --lblk` by `--blk-names` / `--blk-names-exclude` / `--blk-serials`; config file manually editable.
- **Identity is serial-first** (lsblk SERIAL → WWN → synthetic fallback), persisted in the DB; restart re-resolves serial → current name, stored name as fallback — kernel renames across reboots are harmless. AIO bdev name `aio_<sanitized_serial>` is stable.
- **Failure parity, control-plane only**: distrib error_* events are already bdev-agnostic; the AIO gap (no `bdev_nvme` IO timeout) is closed by a hung-IO watchdog in DeviceMonitor (qd-sampling + iostat progress, 3×10 s → UNAVAILABLE; ≥2 stalled devices → node auto-restart); device disappearance handled via /blockdevices inventory sweeps; late-event gate, reset, restart, and new-device-from-failed all have mode-aware branches.
- **Safety**: no vfio/uio binds in lblk mode; SPDK gets the placeholder PCI allowlist `0000:00:00.0` (never an empty list — empty means DPDK allow-all on k8s).
- Phase 1 requires journal-on-device (`num_partitions_per_dev == 0`); `--ssd-pcie` restart-time growth is rejected on lblk clusters.
- 93 unit tests + 10 FDB integration tests.

### 2. sys-memory sizing fix (55417369)

`sn configure --lblk` demanded 2 GiB + full device capacity (102 GiB for a 100 GiB node). Now 2 GiB + 0.2 % of capacity. The nvme path accidentally measured zero (capacity read after driver unbind) and is unified on the same formula.

### 3. Zombie-SPDK cleanup + JM-mesh activation gate (6e7fcb23)

Two field findings from the lblk soak campaigns, both mode-independent CP bugs:

- add-node failure cleanup trusted an RPC-socket probe ("SPDK confirmed down") while the container survived with its restart policy, squatting hugepages and starving retries. New agent endpoint `spdk_process_cleanup` (clears restart policy, synchronous verified remove); `_kill_spdk_until_dead` prefers it with legacy fallback.
- retried add-nodes could join with peers missing their `remote_jm_*` controllers, silently degrading JC quartets until the first journal load collapsed them. New `verify_jm_mesh_coverage` + activation gate: fresh activation hard-fails on unrepaired mesh holes; re-activation only warns (recovery with unhealthy nodes must never be blocked). 11 unit tests.

### 4. env_var: SPDK fork image → `ultra:md-journal-latest` (8385dd9c)

Points the default fork image at the md-journal branch (blobstore metadata journal, simplyblock/spdk#61). **Flagged for reviewer decision**: merging this changes the default SPDK image for every deployment from this branch — drop this commit if main should stay on `ultra:main-latest` until spdk#61 lands.

### Validation

- Unit tier: 1121 passed, ruff clean (post-rebase; the rebase surfaced one semantic conflict — main's cpu-alloc refactor removed a `constants` import `device_controller.py` now needs — fixed in the feature commit).
- E2E: three AWS campaigns on 6-node EBS-only lblk clusters (2+2, ha-jm 4). Final run 2026-08-06/07 on this exact code + `ultra:md-journal-latest`: full 60-scenario dual-node-outage churn soak (graceful / forced / container_kill / host_reboot / 65 s network partitions × unrelated / P-T / P-S pairs) — **60/60 passed, zero fio faults, zero cluster suspensions**.


simplyblock_core/cluster_ops.py
```diff
@@ -256,6 +256,17 @@ def parse_protocols(input_str: str):
...
 
+def _validated_device_mode(device_mode) -> str:
+    """Normalize/validate the cluster device mode ("nvme" | "lblk").
+    Deploy-time only, like enable_failure_domain."""
+    mode = (device_mode or constants.DEVICE_MODE_NVME).lower()
+    if mode not in (constants.DEVICE_MODE_NVME, constants.DEVICE_MODE_LBLK):
+        raise ValueError(
+            f"invalid device_mode {device_mode!r}; must be "
+            f"'{constants.DEVICE_MODE_NVME}' or '{constants.DEVICE_MODE_LBLK}'")
+    return mode
```
mxsrc, line 277: This can be a type annotation.

simplyblock_core/utils/__init__.py
```diff
@@ -1409,6 +1396,107 @@ def addNvmeDevices(rpc_client, snode, devs):
...
+    """Match the node's configured lblk devices against the live host
+    inventory, SERIAL-FIRST: kernel names shift across reboots, so the stored
+    name is only a fallback for devices without a resolvable serial. Returns
+    ``(resolved, missing)`` where resolved entries carry the CURRENT
+    name/path/by-id."""
+    by_serial = {d["serial"]: d for d in host_devices}
+    by_name = {d["name"]: d for d in host_devices}
+    resolved, missing = [], []
+    for entry in configured_entries:
+        live = by_serial.get(entry.get("serial")) or by_name.get(entry.get("name"))
```
mxsrc: This will resolve the wrong device just based on the slot. If a disk is changed in the same slot, the serial will not match, but the name will (e.g. `sdb`). The device will not be flagged as missing, and worse, it will silently use an unrelated device.

Suggested fix: Instead, if a serial is present in the device we're looking for, it _must_ match, only if it is missing we fall back to name-based resolution, and we need to make sure that this is practically only used for initial discovery (i.e. we find it be name, and based on that determine the serial that will be used for future lookups).

simplyblock_core/rpc_client.py
```diff
@@ -1231,6 +1206,36 @@ def bdev_examine(self, name):
...
+    def bdev_aio_rescan(self, name):
+        """Re-read the backing device's size (device grow pickup)."""
+        return self._request("bdev_aio_rescan", {"name": name})
+
+    def bdev_set_qd_sampling_period(self, name, period_us):
+        """Enable queue-depth sampling on a bdev so bdev_get_iostat reports
+        queue_depth/io_time — the hung-IO watchdog's signal for AIO base
+        bdevs (period 0 disables)."""
+        params = {"name": name, "period": period_us}
+        return self._request("bdev_set_qd_sampling_period", params)
```
mxsrc, line 1286: These all silently ignore errors. They should instead use `_request3`, which will raise appropriate errors.

simplyblock_core/rpc_client.py
```diff
@@ -1231,6 +1206,36 @@ def bdev_examine(self, name):
...
+
+    def get_bdevs_2(self, name):
+        """(ret, err) probe variant of bdev_get_bdevs, mirroring
+        bdev_nvme_controller_list_2 — used where the caller must distinguish
+        'bdev gone' from RPC failure without raising."""
+        return self._request2("bdev_get_bdevs", {"name": name})
```
mxsrc, line 1292: This deviates form the original: The warning is not present, its semantics are hidden (e.g. explicitly passing `None` gives the full list, and in the regular use-case (passing a name) its name lies, it does only return one device.

I suggest to instead adapt `RPCClient.get_bdevs` to also use `_request3`, which will silently upgrade it to handle errors, with the added advantage of more robust error reporting elsewhere. This still has the naming issue, but it's well-known in the code base so it can be adapted at a later point.

simplyblock_core/cluster_ops.py
```diff
@@ -1548,6 +1565,27 @@ def _set_node_ana(node_id) -> None:
...
+            if is_fresh_activation:
+                set_cluster_status(cl_id, ols_status)
+                raise ValueError(
+                    "Failed to activate cluster: JM mesh coverage incomplete "
+                    "(journal quorum would silently run degraded): "
+                    + "; ".join(jm_problems))
+            logger.warning(
+                "JM mesh coverage incomplete on re-activation (continuing — "
+                "recovery path): %s", "; ".join(jm_problems))
+
```
mxsrc, line 1645: This runs after lvstore creation, QoS, JC-compression, and MCP narrowing. A failure here only rolls back the cluster status but leaves the other state in place.

simplyblock_web/node_utils.py
```diff
@@ -148,6 +148,184 @@ def get_spdk_devices():
...
+        with open(path, "r") as f:
+            return f.read().strip()
+    except OSError:
+        return ""
+
+
+def _disk_holders(name: str) -> List[str]:
+    """Union of /sys/block/<d>/holders and every partition's holders —
+    catches LVM PVs, md members and dm-crypt without a mountpoint."""
+    import os
```
mxsrc, line 162: Should be top-level

simplyblock_web/node_utils.py
```diff
@@ -148,6 +148,184 @@ def get_spdk_devices():
     return []
 
 
+def _read_sysfs(path: str) -> str:
```
mxsrc, line 151: This should return an `Optional[str]` with `None` in the failure case.

simplyblock_web/node_utils.py
```diff
@@ -148,6 +148,184 @@ def get_spdk_devices():
...
+                    pass
+    except OSError:
+        pass
+    return sorted(set(holders))
+
+
+def _disk_by_id_path(name: str) -> str:
+    """Preferred stable /dev/disk/by-id symlink for a whole disk: wwn-* first,
+    then any other non-partition link. Empty when none exists."""
+    import os
```
mxsrc, line 184: Runtime imports should be avoided, this is better placed at the top.

mxsrc (changes requested):
Looks largely good, most of my comments are very narrow. The one big thing is the device binding issue.

A few things that should be applied globally:
- avoid `except Exception`: This hides programming errors, use specific exceptions if at all possible. E.g. for RPC interaction `RPCException` can be used.
- imports should be at the top of the file. Otherwise modules are loaded at each execution of the function.

PR #1236, opened 2026-08-12
Introduce metrics support to API v2

mxsrc:
The metrics that are exported for Prometheus were missing from API v2. This PR introduces them. Because of some issues with the v1 metrics, they are iterated on and both versions are not compatible:
- Dropped `size_util`/`size_prov_util`: v1's were lossy (`int(used/total*100)`), making them coarse for graphs. Dividing the byte gauges is exact.
- Status as `{status="degraded"}` label instead of numeric code. The label doesn't hardcode magic numbers and survives map changes.
- `health_check` absent instead of `NaN`: Same information, no float sentinel for consumers to special-case.
- Dropped `date`, `record_duration`, `record_start_time`, `record_end_time`: Metadata, the `record_*` trio was never populated by any collector.
- `simplyblock_` namespace: Avoids collision with the node and foundationdb jobs and allows coexistence with v1 metrics.
- Added pool_name to volume series: Preserves the human-readable $pool dropdown that v1 got for free from its (otherwise buggy) label.

Because there is no upgrade path for the prometheus config this will require manual intervention on upgrade. I'll do a follow up PR on the operator to point the helm charts in the right direction.

Part of the changes moves the CPU stats collection to the monitoring service, s.t. the endpoint only interacts with foundationdb.

### DIFF
--- a/simplyblock_core/models/stats.py
+++ b/simplyblock_core/models/stats.py
@@ -1,6 +1,7 @@
 # coding=utf-8
 import json
 import uuid
+from typing import List, TypedDict
 
 from simplyblock_core.models.base_model import BaseModel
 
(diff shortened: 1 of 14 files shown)


Hamdy-khader (commented):
Looks good

PR #1245, opened 2026-08-13
fix(alerting): use ratio query for cluster capacity alerts to preserv…

geoffrey1330:
…e cluster label in fired alerts

### Problem
`Cluster_absolute_capacity_reached` and `Cluster_provisioned_capacity_reached` alerts
were firing with `Cluster [no value] absolute/provisioned capacity reached` in the summary.

The rules used a multi-query approach (two Prometheus queries + a Grafana Math expression)
to compare utilization against the critical threshold. Grafana's Math expression node strips
labels from the result series, so `$labels.cluster` had no value when the alert fired.

### Fix
Replaced the two-query + Math expression chain with a single PromQL ratio query:

- `cluster_size_util / cluster_cap_crit`
- `cluster_size_prov_util / cluster_prov_cap_crit`

The threshold condition fires when the ratio is ≥ 1 (utilization has reached the critical
level). Because there is no intermediate Math node, the `cluster` label is preserved on the
result series and resolves correctly in alert summaries.

### DIFF
--- a/simplyblock_core/scripts/alerting/alert_rules.yaml
+++ b/simplyblock_core/scripts/alerting/alert_rules.yaml
@@ -437,8 +437,8 @@ groups:
               datasourceUid: PBFA97CFB590B2093
               model:
                 disableTextWrap: false
-                editorMode: builder
-                expr: cluster_prov_cap_crit
+                editorMode: code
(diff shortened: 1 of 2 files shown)


Hamdy-khader (approved):
Looks good

boddumanohar (approved):

PR #1246, opened 2026-08-14
Fix node-removal state handling and pause/resume orchestration

Hamdy-khader:
### Summary
This PR fixes storage-node removal state handling during node drain/removal workflows and makes the removal lifecycle explicit and resumable.

### What changed

- Added explicit node states for removal progress:
    - pending_removal
    - in_removal

- Updated distribution/status reporting to treat these states as non-schedulable/unreachable so they are not counted as healthy cluster members.

- Updated node-removal orchestration to:

    - mark a node as pending_removal before the removal task becomes active 
    - transition to in_removal once the orchestration begins
    - resume cleanly on retries without re-running completed phases
     - keep “retry later” behavior when peer nodes are still unavailable or migration work is incomplete
 

-  Preserved correct phase ordering for:

    - graceful shutdown
    - replica relocation
    - finalization
    - device decommissioning

### DIFF
--- a/simplyblock_core/distr_controller.py
+++ b/simplyblock_core/distr_controller.py
@@ -63,7 +63,7 @@ def _persist_target_device_event(device, status, target_node):
 def send_node_status_event(node, node_status, target_node=None):
     db_controller = DBController()
     node_id = node.get_id()
-    if node_status == StorageNode.STATUS_SCHEDULABLE:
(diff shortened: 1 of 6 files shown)


mxsrc (approved):

Hamdy-khader:
Merged as part of PR#1280

PR #1247, opened 2026-08-14
Implemented fdbbackup runner abstraction which can be backed by Docker or Kubernetes

noctarius:
The current task runner to create FoundationDB backups directly uses the Docker client to run the fdbbackup command which fails on Kubernetes.

This PR adds a simple abstraction which is either backed by Docker (default) or Kubernetes (env var `FDB_BACKUP_MODE=kubernetes`).

### DIFF
--- a/simplyblock_core/constants.py
+++ b/simplyblock_core/constants.py
@@ -102,6 +102,17 @@ GRAYLOG_CHECK_INTERVAL_SEC = 60
 # (run 2026-07-21).
 FDB_CLEANUP_INTERVAL_SEC = 60 * 60
 
+# Deployment mode of the FDB backup agent. Set to 'kubernetes' to run the
+# fdbbackup/fdbrestore/fdbcli commands in the agent pod through the kubernetes
+# API; anything else falls back to the management node's docker daemon.
+FDB_BACKUP_MODE = os.getenv('FDB_BACKUP_MODE', '')
+
+# Timeout for a single command executed inside the FDB backup agent
+# (fdbbackup/fdbrestore/fdbcli). `fdbbackup start -w` blocks until the backup
+# completed, so this has to cover a full backup run - the docker client's
+# 60s default silently aborted long backups with a read timeout.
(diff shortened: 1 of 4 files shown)


simplyblock_core/controllers/fdb_backup_controller.py
```diff
@@ -5,70 +5,82 @@
...
 
-def __get_fdb_cont():
-    snode = db_controller.get_mgmt_nodes()[0]
-    if not snode:
-        return
-    node_docker = docker.DockerClient(base_url=f"tcp://{snode.docker_ip_port}", version="auto")
-    for container in node_docker.containers.list():
-        if container.name.startswith("app_fdb-backup-agent"): # type: ignore[union-attr]
-            return container
+def __get_fdb_agent() -> FdbAgent:
```
mxsrc, line 23: Double underscore prefixes are reserved for language builtins. ~~Also I'd rather move it into the client defition itself, following the KMS pattern.~~ There is already a helper in the client definition, I'd rather use that, and deduplicate the helper.

simplyblock_core/constants.py
```diff
@@ -102,6 +102,17 @@ def get_config_var(name, default=None):
 # (run 2026-07-21).
 FDB_CLEANUP_INTERVAL_SEC = 60 * 60
 
+# Deployment mode of the FDB backup agent. Set to 'kubernetes' to run the
+# fdbbackup/fdbrestore/fdbcli commands in the agent pod through the kubernetes
+# API; anything else falls back to the management node's docker daemon.
+FDB_BACKUP_MODE = os.getenv('FDB_BACKUP_MODE', '')
```
mxsrc, line 108: Can't we use `Cluster.mode` to determine this instead of introducing a new value? Otherwise, the default should be `docker` I suppose, at the moment the value is used as `kubernetes` or anything else, which will probably be fine but not explicit.

mxsrc (changes requested):
Looks mostly good, a few nitpicks.

mxsrc:
Closed after discussion with Chris. Superseded by FDB-operator based backups on the kubernetes side.

